PR #2, opened 2017-01-03
do not try to compile nonexistent acid.f90; fixes #1

### LINKED ISSUE #1: CMake error in master - does not compile
I get a CMake error in master after a fresh clone - it does not find acid.f90 - we need to add it or modify the cmake files such that it does not search for acid.f90 any more - @bast should probably know about this

bast:


bast:
@heikef assigned to review this pull request.

heikef:
Looks ok.

heikef (commented):
deleting acid.f90 should work

PR #3, opened 2017-01-04
included acid.f90

heikef:
included acid.f90 and modified CMakeList.txt - no remaining compilation problem - @bast can review


bast:
Thanks!

PR #4, opened 2017-01-06
included g092gimic and lsdalton2gimic scripts

heikef:
@bast could you please check - thank you


bast:
I guess at this point I am expected to just accept the changes, right? Or do you want feedback on the coding? Because my feedback for the future would be to not use classes/objects but build the code from functions and immutable data structures. But it might be now too late for that kind of feedback.

heikef:
yes

bast:
In this case I click "merge".

heikef:
I guess this is ok.

PR #6, opened 2017-01-09
included qchem + fermion++ scripts 

heikef:
The scripts were provided by J. Kussmann - I included this into the README file @bast


bast:
Thanks!

bast:
By accepting these we assume that these sources are provided to us under our license. We should probably state it somewhere.

heikef:
We did not talk about a license to somebody

bast:
I am editing your reply (please do not quote my text if you reply directly through email).

bast:
We simply cannot accept source code if it conflicts with our license terms.

heikef:
Agreed.

PR #8, opened 2017-01-10
Test script updates

mariavd:
Here is the new test molecule C4H4 together with the updated version of the script gimic-test.sh 
Please check the gimic-test-README file for details on how to use it.


bast:
One observation is that the PR contains a commit and its revert. It also contains a merge. Normally I would now insist on cleaning this up but I don't want to be too difficult either. The PR should only contain the relevant commits and the commits should be meaningful (cherry-pickable) units.

mariavd:
I did `git revert 8fbcc9a`. What should I have done instead? I would rather build good habits from the beginning, so it doesn't matter if it is difficult.

bast:
`git revert` is the right solution to undo commits on public branches that other people depend on. In this case you should probably delete the commit with `git reset` (first practice `git reset` in a safe place to not throw away your work). Let us not worry about the merge commit now, we can discuss that another time.

PR #17, opened 2017-01-13
A new attempt for PR for the C4H4 test molecule and the updated test script

mariavd:
I think history has been cleaned now. I'm sorry if there is still something wrong. I will try to fix it if necessary. 

Thank you, @bast


bast:
Yup - looks much better now. Thanks!

PR #20, opened 2017-01-18
Visualizations - fixed NaN in jvec.vti

mariavd:
There used to be a NaN in the jvec.vti file when doing current density calculation in 2D. The reason was that the spacing value for the z coordinate was calculated as 0 divided by a number. It has been fixed to print just 0 for that value. 

Please let me know if the history is again too messy.


bast:
Well. Yes :-) Any good reason why this should not be just one commit?

mariavd:
There is a bad reason. I wasn't able to get them rebased properly. The commits that are related to the NaN fix are mixed with Heike's commits which I had pulled from master, and with some from the test-script branch that I merged with. Now when I do `git rebase -i HEAD~2` there are 9 commits to choose from.

```
  1 pick f74930f Fixed the NaN in the 2D vti files
  2 pick c4017e1 Combining two similar commits for the vtkplot.f90
  3 pick 5ad378b Fixed NaN in 2D jvec.vti files The gimic-test.sh script has been updated to run a test for C4H4
  4 pick 29d1e1c The gimic-test.sh script runs from the test directory rather than from ./test/benzene/
  5 pick 0bb8142 added qchem + fermion++ script
  6 pick 1554bd4 removed swap file
  7 pick 3acb110 The README file contains some details on how to use the provided ParaView state file.
  8 pick 44b2ba2 Debugging src/fgimic/vtkplot.f90
  9 pick 5a11599 Fixed the NaN in the 2D vti files
 10 
```

I tried sqashing those that are related to NaN and dropping the others, but there are issues. In `git status`:

```
Unmerged paths:
  (use "git reset HEAD <file>..." to unstage)
  (use "git add <file>..." to mark resolution)

	both modified:   ../src/fgimic/vtkplot.f90
	both added:      C4H4/int/gimic.inp
	both modified:   gimic-test.sh
```

So then `git add gimic-test.sh` and `git add C4H4/int/gimic.inp` are fine. But `git add ../src/fgimic/vtkplot.f90` can't be merged. 

And ` git checkout -- ../src/fgimic/vtkplot.f90`
error: path 'src/fgimic/vtkplot.f90' is unmerged

I'm sorry for this stupid question. Could you tell me how to fix it?

bast:
First make sure your forked master is exactly as the central master. Then create a new feature branch. Then cherry-pick the interesting commit(s) from the ugly branch to the new branch. Then submit new PR.

PR #21, opened 2017-01-18
Test scripts for the 2D and 3D current density calculations

mariavd:
There are two scripts in the test/gimic-test-2D.sh and test/gimic-test-3D.sh, which can the example for current density calculations in 2D or 3D for benzene. The vti and cube files are compared for differences.


bast:
And now rebase your branch on top of master to get rid of the two merge commits.

mariavd:
Any better now?

bast:
Except that there is the "new try add bug fix ubuntu 16.10 virtualbox" which I think does not belong there.

mariavd:
Maybe now it's fine?

bast:
A lot better now. When submitting a PR please always check whether there are commits in there which you did not expect to be there.

PR #22, opened 2017-01-18
Fixed the NaN in jvec.vti 

mariavd:
Just one commit pushed this time with the fix in the vtkplot.f90 source file.


bast:
Super!

PR #24, opened 2017-01-20
Vectors in jvec.txt, jvec.vti and jmod.txt are no longer converted

mariavd:
Vincent's suggestion was that it is redundant to convert the vectors to atomic units. As far as I can see no side effects to the code have appeared but it would be nice for others to verify, too. 

Note: the two other commits are meant to be included to the master but I'm not sure it is a good idea to combine all changes into one commit.


heikef:
Looks ok. Thank you.

PR #27, opened 2017-02-07
Corrected the 2D and 3D tests

mariavd:
The scripts for the 2D and 3D tests should now be fixed. Please let me know if there are any issues.


bast:
Super! Thanks!

mariavd:
To summarise what I did:

There are reference files called *.ref. Running the test script creates *.vti, *.cube and *.txt, so it avoid overwriting those that were previously on the master, I deleted the  *.vti, *.cube and *.txt from the branch. There are also the Paraview pvsm files that Heike had provided. There was an issue that they need the local path to the vti files, so in the Paraview-README file there are a couple of one-liners to fix the path.

heikef:
Great! I was reading and then the merge has already been approved. -)) It would be cool if things could always be that fast.

PR #30, opened 2017-02-08
adapted cmake files and inputs for test.py

heikef:
adapted gimic/test CMakeLists.txt and test.py as well as *.inp.py files such that one can run now "make test" and bonintegral and vectors will be executed 
this needs to be adapted further for the test we already have - as I understood Maria will prepare this - a working test.py might be helpful in this respect


bast:
Thanks!

PR #34, opened 2017-03-06
added B field information in cdens output

heikef:
Included B field information when 3D calculation is run. Changed mol and xdens files for test suite.


bast:
Looks like two changes are in one single commit. I will integrate it now but next time please try to keep things separate.

Issue #31, opened 2017-02-24
Migrate documentation to Sphinx/RST


bast:
We need to migrate away from https://github.com/qmcurrents/gimic/tree/master/Documentation.

bast:
Done. Needs some cleanup of course.

PR #41, opened 2017-06-27
mv doc from tex/pdf to rst/sphinx

bast:
Documentation is now here (after accepting this PR): http://gimic.readthedocs.io


bast:
This fixes #31.

PR #47, opened 2017-06-28
Tests are now run with "make test"

bast:
Plus other clean-up in testing.


heikef:
The outcommented code is due to an attempt 1) to interface gimic also to London and 2) convert to pygimic. However, that has never been finished (as far as I know) so I think it is ok to remove it.

bast:
Attempts belong to branches :-) Out-commented code is simply confusing for everybody else (is this working? is it broken? is it needed? is it not?) so I remove it whenever I see it :-)

Issue #45, opened 2017-06-28
Go through current doc and remove obviously outdated information


bast:
@heikef please go through http://gimic.readthedocs.io and please remove/update under doc/ everything that is obviously wrong or outdated. Since you are used to the code you are in a better position to spot outdated information. For me it is very confusing to follow documentation that is partially completely wrong.

heikef:
OK - on my to do list

PR #48, opened 2017-06-29
updated documentation 

heikef:
Outdated information is removed. See issue #45 

An update on how to run calculations using the other scripts for QChem, Fermion++, G09 and LSDalton still needs to be added. I will open a new issue in this respect.

### DIFF
--- a/doc/index.rst
+++ b/doc/index.rst
@@ -1,7 +1,7 @@
 
 
-GIMIC: gauge-including magnetically induced current 
-===================================================
+GIMIC: gauge-including magnetically induced currents 
+====================================================
 
 .. toctree::
    :maxdepth: 2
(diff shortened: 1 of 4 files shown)


bast:
Thanks. Why was the "method" added in gauge-including magnetically induced current method?

heikef:
We always write method or approach in publications. It felt as something would be missing.

bast:
But it breaks the acronym. Is the word "method" important in the title? I find it without "method" catchier.

heikef:
I do not see it as a religion. I will remove it and file a new pull request.

bast:
Neither do I :-) You don't have to close this PR - you can append to this one. Reopening.

heikef:
Ok, thanks for the hint.

bast:
I suggest to change "current" back to the original "currents".

bast:
Thank you.

heikef:
The official title included the "s" if I may say so.

Issue #51, opened 2017-06-29
Update example directory


bast:
@heikef agreed to update this based on one of our tests.

PR #52, opened 2017-06-29
added the benzene test example into examples with reference files

heikef:
This is related to issue #51 and based on the input files and date we have in /test


bast:
Super. Only modification I would like you to make is to modify the commit message to auto-close issue #51.

heikef:
OK - next time I will add auto-close and then the number of the issue. Thanks for the hint!

bast:
OK but please try to fix this one (for the exercise).

bast:
```
# edit last commit message
git commit --amend

# never ever push --force to the central repo, only to your fork and only the feature branch
git push --force origin mybranch
```

heikef:
Thank you. I followed your instructions. Please check if this is now acceptable.

bast:
Please follow https://help.github.com/articles/closing-issues-via-commit-messages/.

bast:
Use "closes" or "fixes".

heikef:
Thank you for the link! I understand. One needs to use the right keyword. I think it should work now.

bast:
Sorry for being pedantic but leave out the "issue" ("fixes #N" is enough). Also this should be part of the first commit line so that it is visible in web overviews.

heikef:
Ok - I modified the commit message. It starts now with "closes #51". There is not much left what could be wrong this time. Thanks for being patient!

bast:
And now the finale: please both the message and the closes on one line. And the line should not be longer than approx 50 characters. Be brief and clear. If more info is needed, add a paragraph below.

bast:
This is the reason: https://github.com/qmcurrents/gimic/commits/master - I need to see what is going on without needing to click around and inspect diffs.

bast:
:+1:

PR #55, opened 2017-07-10
G09 interface documentation - added input example

heikef:
Tested the g092gimic script and updated documentation. We have now a simple G09 input example for benzene that works for G09 version d.01 and information on how to generated a formatted checkpoint file. The script seems not supportive for open-shell cases. I included this information.


bast:
Thanks!

PR #57, opened 2017-07-18
remove a lot of unused code

bast:
More to come.


bast:
My comment above is misleading - what I wanted to say is that it is ready to be merged but there is a lot more cruft that needs to be removed.

heikef:
The netcdf stuff might be useful at a later stage but for now it is best to clean up.

bast:
Yeah, add it once you need it. Until then it is just weight.

PR #58, opened 2017-07-18
Large refactor under the CMake hood

bast:
Slight changes in the setup script but changes should largely be transparent.


heikef:
Thanks! A lot of cleaning up work has been done. 
Why is the renaming of some files needed? 
It is a good idea to remove what is not needed, but we will need later on automatic BLAS library or LAPACK detection.

bast:
Renaming was to simplify for my own taste - perhaps not really needed. Currently we do not use any BLAS or LAPACK that's why I nuked that code.

bast:
The motivation for all this work was to switch to http://autocmake.readthedocs.io - this should give us a more portable and more maintainable CMake structure.

heikef:
Do we have to rename now all *.f90 files?

heikef:
In any case it is just a small change from *.f90 to *.F90, no problems expected.

bast:
Aha - the renames to F90 were needed because these files use preprocessing and there the standard is to use uppercase suffix. For files which are not preprocessed, standard is lowercase f90.

heikef:
Ok that means we can have both.

bast:
Yes. We need to have both.

PR #59, opened 2017-07-21
updated documentation

heikef:
Updated documentation. The Turbomole part is now a bit clearer and I pointed out on the make test feature, which is new and probably very helpful to check if an installation is working correctly.


bast:
Super!

PR #60, opened 2017-09-04
Feature documentation

heikef:
Added new scripts from Vincent and updated documentation. Note this needs more careful testing but for the time being it is important to share the new scripts with the users.


heikef:
I modified only the scripts. Why are the tests failing? This is not related.

bast:
Test failure is unrelated (please consult logs). Same problem as discussed in #62.

Issue #62, opened 2017-09-04
Automated testing crashes


heikef:
Automated tests crash. We need to figure out why and fix this.

bast:
The reason is `#include "numpy/arrayobject.h"`, see one of the logs under https://travis-ci.org/qmcurrents/gimic. We had this before. I really need to replace the Cython layer. Working on it. Will try to figure out a workaround until then.

heikef:
Thanks for having a look at this. We should share the new open-shell scripts with the user community as soon as possible. Would it be acceptable to merge even when the tests crash? The reason for the crash is not the new modified script. Or would this be considered as bad practice?

bast:
I believe this is fixed since #64.

PR #64, opened 2017-09-12
Fix Travis.

guacke:
We seem to have broken your Travis CI build. We will try to fix this in the following commits.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,7 @@
 language: python
-
+python:
+ - 2.7
+ 
 sudo: false
 
 matrix:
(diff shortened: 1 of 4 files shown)


bast:
I am not sure it was you. There was an issue before. We need to check the log as soon as it completes. It might be an issue on our side.

bast:
OK seems to be a different error but on the upside it looks like you have fixed #62. Let me know if you need assistance fixing it. Thanks again for your work!

guacke:
This commit should fix the Travis CI problems caused by our previous commit.

bast:
Thank you! I will squash the commits into one.

PR #65, opened 2017-09-15
Bug fix: change != to is not and apply dos2unix to files.

guacke:
Fixes minor issues related with g092gimic (DOS format and small error in Python code).


bast:
Thank you!

PR #68, opened 2017-10-18
document how to install on Stallo

bast:


heikef:
Thanks Radovan!

heikef:
Is the setup script now needed for both branches master and stable? I thought it will be replaced at one point.

bast:
No the setup script is here to stay according to my information :-)

Issue #61, opened 2017-09-04
Open-shell test case


heikef:
A test case for open-shell molecules needs to be added. This has to work with the updated G09 scripts provided by Vincent. I have this on my to do list.

bast:
Closed by #71.

PR #71, opened 2017-11-02
Feature open shell

heikef:
Added an open-shell example. This needs to be checked though by the community. Closing #61

### DIFF
--- a/doc/usage.rst
+++ b/doc/usage.rst
@@ -195,6 +195,40 @@ $ Gaussian2gimic.py --input=benzeneg09.fchk
 Note, for open-shell cases you need to add "gfprint pop=regular iop(10/33=2)"
 and use the Gaussian "log" file instead of the "fchk" file. 
 
+Example input for benzene triplet dication 
+
+::
+
+    %LindaWorkers=cib26-2
+    %NProcShared=20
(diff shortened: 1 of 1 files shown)


bast:
Does it mean we are not sure about the correctness? If yes, I would mark it there to avoid possibly confusion on our side where we accept a value as given. I will also squash the merge commit away but this can be done on GitHub side.

heikef:
I think the value is correct. Tested it also on HF level but I never got a reference value from Vincent for this problem nor any other test case for closed and open-shell applications. So I had to set up things on my own. There was a lot of confusion and back and forth emailing with several users regarding open-shell calculations. Therefore, I decided to share on github what we have figured out so far and let it be tested by the user community. That will most likely converge to something reasonable and usable. 
 
Alternatively one would need to recompute published values for fluorinated benzene for example including ECP's. However, I think it is not my job to figure on how things work in Gaussian and to redo others work when those being involved in the publication could easily have a look at our github and contribute.

bast:
OK thanks for explanation. Merging.

heikef:
I know it is not optimal but I think it is better to share what we have.

bast:
I agree.

PR #72, opened 2017-11-08
Feature open shell

heikef:
Updated documentation after getting more information from Vincent. Tested the modified keywords and got the same result as before.

### DIFF
--- a/doc/usage.rst
+++ b/doc/usage.rst
@@ -204,7 +204,7 @@ Example input for benzene triplet dication
     %Chk=benzeneg09.chk
     %mem=2000mb
 
-    #p POP=FULL GFPrint Symmetry=None B3LYP/DEF2TZVP SCF=Tight NMR=GIAO Int=NoBasisTransform IOp(10/33=2)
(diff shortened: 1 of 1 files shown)


bast:
That's really great but please check the commits that are part of this PR. There are four but I think you only wanted to submit one. In other words create new branches for new PRs.

heikef:
It is related to the same topic as the last pull request so I thought it is ok to work on the same branch. OK, I will set up a new branch and file a new pull request.

bast:
New PR: always new branch :-) If you want to modify a PR, then you stay on the same branch.

heikef:
If I stay on the same branch I have to modify the commit otherwise my pull request gets rejected. It is not very convenient if small changes require to set up a new branch all the time when the topic does not change. I filed now a new pull request but created a new branch, because this is faster for me since I know the procedure on how to do it.

bast:
New PR, new branch. Always. Once a PR gets integrated, you can and probably should remove the branch. If you are asked to make changes to a PR that is not yet integrated, you stay on the same branch.

heikef:
Amen. -))

PR #77, opened 2017-11-30
do not look for python libs; fixes #76

### LINKED ISSUE #76: Do we really need to detect PythonLibs?
Detection is a pain since we need to provide Python header path and library location at configuration time and I am not sure these are even used and therefore possibly not needed? It works for me without.

bast:
It seems that the detection is unnecessary complication.


heikef:
Thanks, I think this will make our life much easier.

Issue #84, opened 2017-12-01
Change input format to YAML


heikef:
I got some user feedback regarding the input of gimic and would like to open an issue to discuss on how things should be restructured or not.

bast:
My long term plan is to switch input format to YAML format. I have a related project where we want to develop the next generation input parser and replace libgetkw. All suggestions welcome for features you would like to have implemented there.

bast:
And I think input and scripts is a too wide topic. I think scrips and input format are two different topics.

heikef:
Why YAML? I found this page: http://yaml.org/ Does not look tempting to me. 

Changed the discussion topic for not being too wide. I think this is related to input format plans.

bast:
That page is scary but YAML is the de-factor standard input format. Easy to parse for machines, easy to write for machines, easy to write for humans, readable for humans. You can have nested sections. My favorite format. But don't worry, before going there I would create a proper proposal demonstrating how the new input would look compared to now. What I can already promise is that it will look a lot cleaner and easier to read.

bast:
Most importantly: it is a standard format. Creating own formats is not a good idea.

bast:
Example - instead of:
```
calc=integral          # cdens, integral, edens, divj
#dryrun=off        # don't actually calculate (good for tuning grids, etc.)
title=""
basis="MOL"
xdens="XDENS"
debug=1           # debug print level
openshell=false
# show_axis=true   # mark "up" axis in .xyz files
magnet_axis=X #[-] i,j,k || x,y,z -> align magnet along axis

# grid type can be: base/std, bond or file (see examples below)
Grid(bond) {                    # define grid orthogonal to a bond
    type=gauss                  # gauss distribution of grid points
    bond=[1,2]
    fixpoint=4
    distance=1.32               # place grid 'distance' between atoms
    gauss_order=9               # order for gauss quadrature
    grid_points=[30, 30, 0]
    height=[-5.0, 5.0]
    width=[-2.2, 5.0]
}
```
With YAML:
```
calc: integral          
title: ""
basis: "MOL"
xdens: "XDENS"
debug: 1           
openshell: false
magnet_axis: X 

grid:
  type: gauss                  
  bond: [1, 2]
  fixpoint: 4
  distance: 1.32 
  gauss_order: 9 
  grid_points: [30, 30, 0]
  height: [-5.0, 5.0]
  width: [-2.2, 5.0]
```

heikef:
[...]

PR #88, opened 2017-12-01
More cleanup in testing

bast:
Another quarter million lines of text gone.


heikef:
I assume the "nuclear waste" is only removed regarding the test example. The out commented keywords are important and should be kept as an option for the input. For the specific test case we look at here they are not needed indeed. - Yes, this is the case. -))

bast:
No, outcommented code does not belong here. Keywords may and will change and then the commented code is confusing both for users (why does this not work anymore?) and developers (will this be needed in future? but nobody is using it).

For me outcommented input examples are a symptom for the following frequent problems:
- input format is not intuitive
- keywords are not intuitive
- documentation is not helpful

heikef:
Thanks for the cleanup.

heikef:
I think therefore we have issue  #84. 

From a user's point of view. The keywords are useful and needed. The input format is very intuitive when you understand the "gimic logic". Documentation can always be optimized. At the moment I would not call it unhelpful.  

I want to keep the keywords!

bast:
You want to keep out-commented keywords in the tests? Why not put them in some examples? The tests should test one thing only. I find it very confusing to have them in the tests.

heikef:
No, I do not want to keep out-commented stuff in the tests. I was pointing out that the keywords that are out-commented are still relevant and should not be in general removed from the input. This is not related to this specific test. Apparently that was not clearly stated. Sorry for this. Hope this is clear now.

bast:
OK I was of course not planning at removing any keywords :-) Only removed out-commented lines.

heikef:
Good. -)) Just wanted to be sure about this. One never knows.

PR #95, opened 2017-12-06
added test for keyword magnet

heikef:
Included tests for the keyword "magnet". For simplicity I used the benzene standard example. If this works fine I can add more feature tests for other keywords.


bast:
It looks like MOL and XDENS are included as a real file but perhaps we could use symlinks like in the other tests to save disk space and avoid code repetition and divergence?

heikef:
Good point, after doing a git pull and investigating the test structure I saw that all test directories have MOL and XDENS as files and not as symbolic links. I thought this is ok like this. I will add links and test if everything runs through.

bast:
That is surprising. When I clone the repo, they are symlinks.

heikef:
Sorry, my mistake I was in the build directory. Everything is soft linked in the test directory.

bast:
But the big files are still part of this PR and would become part of the history. Squash the two commits together.

heikef:
I deleted them. Why is this a problem. They are away now.

bast:
You have one commit adding them, one commit deleting them. Would be better to have one commit that neither adds them nor deletes them, but only adds symlinks.

heikef:
I know, but since I never managed to squash I had to figure out on how that works and do it by try and error. The last change should be fine. Only one commit line with a clear message.

bast:
What are we exactly testing here? What numbers do we precisely compare against reference values? From the output? Or from some generated files?

heikef:
We compare against the files collected in the reference directory. We are testing the keyword "magnet", which can be used instead of the magnet_axis=X option. This is an important keyword since one can define a precise direction of the external magnetic field. In the present test example exactly the same output as obtained with the X option should be obtained for the chosen B field direction defined via the other keyword.

bast:
So what this keyword does is changing the orientation of the magnetic field?

heikef:
You do not change the B-field - you define the direction you want. Gimic does not have a default direction for B, which is very nice because it adds flexibility.

bast:
OK I will rephrase: So what the magnet keyword does is to define the orientation of the magnetic field?

heikef:
My plan is to add step-wise also other keyword tests to cover ideally (almost) all keyword options we have.

heikef:
Yes.

heikef:
Just to be clear we NEED this keyword in Gimic. -))

bast:
Look, I am not trying to remove any keyword, I am only trying to understand what the keyword does. The name is not intuitive but that is another story for another PR some time later. The reason why I am asking this is that this PR adds a lot of new files that take time to test and maintain and I am asking myself whether we could test this feature with less files and a faster test.

heikef:
When we want to have a test for each keyword we need to add much more files. Gimic has the integral "intgrl" option and 3D "cdens" option. So I added tests for both including 2D, which is a subcase of 3D. Of course I can remove the 2D option to reduce data. About the other two options I am not so sure. We could agree to use only the intgrl option for testing for example.

heikef:
For me the name is meanwhile intuitive but that can also be due to the fact that I am using the program rather long. Concerning the removal issue - I prefer to tell things from the very beginning (perhaps also due to some previous experience not related with you). When a programmer gets into the "clean up modus" it is better to be careful. ;)

heikef:
I think it is clear by now what the keyword does? - It defines B_x, B_y, B_z as "magnet=[B_x,B_y,B_z] plus and minus signs are allowed adding more flexibility, which is a good thing in my opinion.

bast:
You mentioned that there is no default in GIMIC so what happens if I leave that keyword out? I am really not trying to remove any functionality and if that moment comes, then not without a discussion first and the discussion would not happen in this pull request. Here, all focus is on the testing.

heikef:
If you leave this out the program stops and asks you to give a B field.

bast:
So the thing I don't understand then is this: we have a couple of tests and a couple of tests with the magnet  label and keyword. If there is no default and the program stops, what are we testing in the tests without the magnet label/keyword?

heikef:
There are two keywords one is magnet  - the other one is magnet_axis=X or =x,y,z . We are currently only testing the X option, in this option the program automatically calculates the optimal direction of the B field perpendicular to eg. the bond you are looking at or relative to the chosen grid in the cdens option. (see one of my comments above) I thought it is good to test both keywords. Therefore I added this extra test.

heikef:
I know for a new user this sounds confusing but the different options are really useful when doing applications on different kinds of molecules, being not necessarily planar.

bast:
OK now I understand. The question is: do we get more test coverage with all the new files or do we get the same test coverage just with one? Depends how modular the code is. Better safe than sorry in doubt.

heikef:
If we want to be on the save side we better test both options. I mean intgrl and cdens for the keyword. Otherwise only intgrl might also work for the moment if we want to save time.

bast:
We want to test the two options. The question is whether we better need all 3d, 2d, and integration and all files to test the two options. If yes, then yes. Just want us to be sure.

heikef:
We could skip 2D. The results must be the same in the case of benzene as those obtained with the X option we already have. Perhaps that also saves files?

bast:
Would it be enough to check the integrated result? Do you think the 2D/3D plot can be wrong if it is correct without magnet and if the integration is correct with magnet?

heikef:
Ideally yes, but as I said intgrl and cdens are two different parts of the program with different grids and so on. If we want to be on the save side we add both. It does not take that long either.

bast:
Yes in this case we want to test both. So do we have a "minimum" (kind-of) set of files to cover all the code parts that we want to test in this PR?

heikef:
As I said we could skip 2D, but it does not hurt to keep it. I think at this point and for this keyword it is not necessary to be "greedy". Most of the keywords that could also be tested are related to intgrl. So the number of files we have should ideally not explode.

bast:
Then let's go for it.

heikef:
Great!

PR #99, opened 2017-12-07
compile and test on two cores

bast:
Probably/hopefully speeds things up a bit.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -25,9 +25,9 @@ install:
 script:
   - ./setup
   - cd build
-  - make
+  - make -j2
   - make install
-  - make test
+  - ctest -j2
 
 notifications:
   email: false


bast:
It speeds up things a bit but also is IMO a good test that we got dependencies right when compiling on multiple cores.

heikef:
Is this really needed? Normally I am using something like make -j20 when compiling. Do we really need to add the -j2 option into the script? It is not a big thing, I am just curious.

bast:
It's faster and we might detect bad CMake code that may happily work on one core.

bast:
But it's not really really needed :-) We can live happily without if there is a good reason to not have this.

heikef:
Ok, ok, we do everything to improve the code. Agreed.

PR #100, opened 2017-12-07
added test for keyword rotation

heikef:
The rotation keyword works for the ingrl option. You define three angles. In this way the integration plane can be rotated and adjusted as the user wants to. In some cases this keyword is very useful. In the present test example I put the plane such that it does not cross the C-C bond of benzene but goes directly through the CH bond.


bast:
Have you also tried that the test breaks if you change some input or change the reference result slightly?

heikef:
The script takes out the number from the reference file and compares. I do not see why I should modify this. But I tried a modification in the input file and then it crashed. I modified the grid keyword and then the test failed.

bast:
Please verify that the test really works by observing it failing. Change some tiny digits in the reference and observe that it really fails then. Too often I have seen tests that happily passed always.

heikef:
Ok, I understand your point and tested this. I modified the third digit in the reference file and the test just failed as it should. Is this acceptable?

bast:
Very good!

Issue #94, opened 2017-12-06
Add more tests to cover functionality


heikef:
We need to add more tests ideally for each input keyword we have. In principle we can stick with the benzene example.

mariavd:
What kind of tests? I can prepare files but what do you mean exactly? Are there any keywords that are missing in the three test input files that we currently have?

bast:
I think this issue is too broad. We need more specific issues, then it is also clear what to do and when the issue is done. Alternatively we should list here what needs to be done for this issue to be resolved.

heikef:
The following keywords could be tested: - all for intgrl - module
grid(bond) = even
spacing
rotation
radius - this one should also be checked with cdens


I will open another issue for more general tests that should be added.

PR #103, opened 2017-12-07
added test for grid bond even

heikef:
Normally one uses grid bond gauss but one can also use the "even" keyword. This is not very much used but since it is available I added a small test example. Note, cdens with grid even is a lot used so it does not harm to test it here as well. Closes issue #94


bast:
Awesome!

PR #104, opened 2017-12-08
added test for read in external grid file closes #97

### LINKED ISSUE #97: test read in external grid file
A test where an external grid file is read in and processed should be added.

heikef:
I would like to add this test such that we have an example on how to read in an external grid file. While working on this I realized that it only works for cdens. Making in work for intgrl could be a separate issue. Note, the grid file is not very good and there is room for improvement, which can be done later. For the moment - recalling a meeting with @bast - this option should be expanded anyway. So I think this could be a start from which we could work on improving things.


heikef:
Strange on my machine all tests passed.

bast:
I think the external grid can be made much smaller. This can be tested on a tiny grid IMO.

heikef:
Agreed. I just took a gridfile that was available to share the input, because this input information is important. The gridfile required at the moment for Gimic is for example different to those you provided. Here we read in only the grid point, while in your MY_GRID files also information about the weights for each point are given. I think the integer numbers refer to the shell but this is another topic and need not be discussed here. 
I will modify the gridfile, I think a meaningful grid is not that important here. So my plan is to go for the fast option. If we want to we can change it later. 
Alternatively, I can only add the input to the documentation and do not add a test about this at all.

bast:
I think it is a very good idea to add the test but this can be tested with a grid file containing as few as 10 points or so.

heikef:
OK. Done.

bast:
Thank you :-) but you know me already since I will now ask you to put the two commits together so that we don't add 200k lines in one commit and remove them again right after.

heikef:
Had initially 4 commits and reduced them to the two you see here. The conversation is a bit confusing when at the end only one commit is left, but ok I see your point and will neither be greedy nor start another discussion about this.

bast:
Great work!

PR #107, opened 2017-12-12
added giao=off test

heikef:
Added a test where the keyword giao is switched off. Found also a bug. There are two arrays that do not get allocated when giao=off but they are used later on. Corrected this. Tested with giao=on/off and the values look ok, so I feel confident to share this but would appreciate if someone else could double check. One never knows....


bast:
Sorry there is now a conflict due to the other PRs. Can you please fix that?

bast:
Sorry for being a bit tough here but please also describe the change in the commit message. Reading the commit message one could get the impression that you only added a test but you also changed the code.

heikef:
I explained above what I changed. It is a bit long for a commit message.

bast:
Sure but what if 10 years from now we are not on GitHub? Then all that counts are commit messages. There is no practical space limitation for commit messages.

heikef:
Done. Albeit I have some doubts that someone will use this code after 10 years....

bast:
Oh it will be used in 10 years and people will have a chuckle over this discussion :-)

bast:
Something is weird with these commits: https://github.com/qmcurrents/gimic/pull/107/commits - I am not sure what you did but do you see what I mean? Please compare commit messages to commit diffs.

heikef:
Sorry, but I do not see what you mean. I changed the commit message as you requested it. There were several commits before so I squashed them together because I knew you would complain about this. Now we have one commit message with all information we can possibly pack into it. I do not see what is considered as wired now nor what am I supposed to do.

heikef:
In my "git log" the first commit message "added giao=off test" does not appear any more.

bast:
OK now everything is squashed to one commit - fine. What would have been a bit nicer is to have the bugfix in one commit and the new test in a separate commit. Also "added test for functionality X" would have been a more useful commit message compared to "modified CMakeListst.txt" - I can see from the log what was modified but it is not clear from the commit message *why* it was modified. But I will merge it now and let's do it better in future.

heikef:
I was working on adding the new test. Then it crashed and I had to localize and eliminate the bug. This is the reason why the commit order/logic has not been optimal. It is also hard to estimate when only one commit message is allowed and when two commits are allowed.

bast:
That's why we use Git where you can edit, split, combine, reorder your commits later.

heikef:
It just takes ages to get a pull request approved. However, I am learning. -))

bast:
If you want I can just merge them always. It also takes my time to review the stuff and the goal is to get better code, more learning, and more understandable history.

heikef:
I prefer to see you happy. ;)

PR #116, opened 2017-12-29
removed not used subroutines 

heikef:
Removed subroutines such as contract2, ctensor2, jtensor2, divergence2 because they are not used. All test went through. I think when these routines are not used we do not need to keep them.

### DIFF
--- a/src/libgimic/divj.f90
+++ b/src/libgimic/divj.f90
@@ -59,35 +59,6 @@ contains
         div=djx+djy+djz
     end function
 
-    function divergence2(rr, bb) result(div)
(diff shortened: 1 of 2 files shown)


bast:
Removed code is debugged code. Excellent.

heikef:
I just hope we do not regret it at one point but for the moment it is not needed. Thanks for the merge.

bast:
We have the Git history. Nothing gets lost forever.

heikef:
Sure, but we need to remember where to search and what to search.

bast:
That's why we need good commit messages.

heikef:
Indeed. -))

PR #121, opened 2018-01-10
Rotate around midpoint rather than the origin of the coordinate system

mariavd:
Lukas and I were working on the rotation keyword. It used to rotate the integration plane around the origin of the coordinate system but we changed it to rotate around the midpoint of the defined bond.


mariavd:
I am investigating the failed test:
```2/17 Test  #3: benzene/vectors ```

lnw:
The failing test 2/17 is because there is no useful reference point around which to rotate defined in an even grid.  Given that in this case one manually defines an origin and two basis vectors that span the grid I cannot see a reason to apply a rotation.  I suggest to disable any rotation in combination with even grids.

mariavd:
The test for the keyword rotation is now passing successfully with the new stdout reference file for it.

heikef:
All checks failed sorry I can not merge. What is the motivation for the changes you introduced?

mariavd:
Only two tests were failing, not all of them. The first one was the test for the keyword rotation, which apparently needed new reference values to compare with. The second failing test is the one that Lukas commented about.

heikef:
You have to correct then the check that failed. Please also explain why you would like to add the changes.

bast:
Only test benzene/vectors is failing at this moment.

lnw:
The idea behind the changes is, that a rotation always has a reference points around which the rotation is performed.  Until now, the reference point was the origin of the coordinate system.  We think it makes much more sense (and allows for more intuitive application) to use the point at which the integration plane intersects the bond.
Put differently, a single rotation is not enough to map the initial plane to any place+orientation in space.  For that, shifting is required.  So, given that the operation can only access a subspace, it should be the most useful subspace.

bast:
Is there a use case for rotating the plane when you don't want to rotate around a bond?

lnw:
@bast I'm not sure if I understand your question correctly.  I think it is only ever useful to rotate the plane around a point that lies somewhere on the bond (that's why I think the change is useful).  Furthermore, I think when one defines the plane explicitly by giving its origin and base vectors, then every plane can be defined and no rotation is ever needed.

heikef:
@lnw I assume an "angle integration" would then still be feasible? Why should the plane only be rotated around a point on the bond?

lnw:
I'm not sure what you mean by angle integration.  I'll try to draw a picture some time, I think that will be more efficient.

heikef:
@lnw  you rotate around the center of a ring along a bond for example. Is this still possible?

lnw:
@heikef  Maria and I added the option to define the origin of the rotation, so you can choose the centre of the ring and use that to scan over that angle.  In particular you can do something similar for rings that are not centred at the origin of the coordinate system.
If the option is omitted, the intersection of grid plane and bond is chosen (which I believe is most often what is desired)

heikef:
@lnw this is great but somehow there is still a test that fails

lnw:
@heikef The second point is was mentioning is the following:
Currently there appear to be four ways in which a grid can be defined (std, file, bond, base).  If I understand correctly std, file, and base allow to specify any grid in space, so any additional rotation seems pointless.  We therefore suggest to ignore the rotation option for all except the bond grid definition.  For that, the test case that is currently failed would be changed (and then work).

heikef:
@lnw the test that failed is "read grid" - test 17

mariavd:
Is Radovan taking care of setting up the prepared test directories or should I figure out how to do it?

heikef:
The new test should be part of this pull request. If you have a look at the test directory you will easily see on how to do it. As I saw you already did it. -))

heikef:
We have 7 commit messages now. Is this really necessary?

bast:
About tests: please add them yourself. To do this `git grep` for an existing test, it will hopefully be evident how it works but if not, please ask questions and I will answer here and by improving docs.

mariavd:
Sorry, I had forgotten to `git add` the test file, too. This explains why Travis only executed 17 tests. 

If you prefer one combined commit instead of the current 8 ones, I can try rebasing them.

heikef:
Yes, we do not need so many commit messages. I think one is sufficient, maximum two.

mariavd:
@heikef Sure, just give me some time... Travis still executed 17 tests. I need to figure out why.

mariavd:
I give up. I don't know what's going on. Calling `./test` in the directory `keyword-rotation_origin` passes but when you run `make test` from the build directory only 17 are detected.

mariavd:
@lnw noticed that the test directory was missing in `test/CMakeLists.txt`. I added the directory name manually to the file and at least locally `make test` now runs 18 tests. I will commit that change and then if it works on Travis, I will rebase the commits into one.

mariavd:
Well, I squashed the 9 commits locally but I am not confident about trial and error how to get it pushed without messing up the upstream history.

bast:
Good you figured out the testing, we need more docs :-) About the squashing: if you are unsure to not break things, create a backup branch from the tip of the branch that you wish to rebase, push the backup branch, then rebase. If things go wrong, you can always go back to the backup branch. If things go right, you can delete it later.

mariavd:
@bast So I should make a new pull request from the new branch that I will create? (but the commits are already rebased on my rotations branch but I cannot `git push` since the local and the upstream have diverged)

bast:
First push your backup branch, then rebase (you have done that), then push with `--force` to this PR branch. No need to open a new PR.

mariavd:
Done. Before accepting the PR, I noticed that in `gimic.in` there are some changes that I have not done. I will restore it.

mariavd:
All good.

heikef:
Thanks!

PR #130, opened 2018-01-22
added formula for GIMIC and ACID into documentation closes #126

### LINKED ISSUE #126: Main equation could/should be part of the web documentation
@heikef it would be wonderful if you could put the main equation for the current density on the web documentation. The documentation understands LaTeX notation. I think this would be really valuable and show what exactly the code is doing.

heikef:


bast:
Wonderful - thanks!

PR #131, opened 2018-01-23
Introducing interactive bash scripts for setting up GIMIC jobs (fixed #108)

### LINKED ISSUE #108: Add new scripts for scans
This is to make usage more convenient for users. @mariavd will take care of this

mariavd:
So, let's see if this branch will be more successful.

### DIFF
--- a/jobscripts/3D-run.sh.in
+++ b/jobscripts/3D-run.sh.in
@@ -1,0 +1,116 @@
+#!/bin/bash
+
+function calcVertices() {
+cat coord | sed -n "2,/^\\$/p" |  awk 'BEGIN{
+  			 infinity=10000;
+			 offset=8;
+                          minX=infinity; 
+                          minY=infinity; 
+                          minZ=infinity;  
+                          maxX=-infinity; 
+                          maxY=-infinity; 
+                          maxZ=-infinity} 
+                          /\$/{next} {
+                                  if ($1 < minX) {minX = $1;}; 
+                                  if ($2 < minY) {minY = $2;};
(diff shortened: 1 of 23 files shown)


bast:
I think it is good to share this and from my perspective I approve. There is one risk though: these scripts might go stale one day. Also one day the interface/keywords in GIMIC might/will change and then these scripts will either not match anymore or somebody has to remember to update them. Another option would be to keep the scripts in a separate repo under the qmcurrents namespace. But also this option carries the same risk that they will go stale. It is actually not risk but a near certainty but still it is better to share than to keep them on a laptop. So take my concerns here with a grain of salt.

mariavd:
Well, once they get old, and if I cannot spend the effort to update them, then they can simply be deleted. I am fine with that. I hope that at least at the moment they will be useful to somebody. 

If you introduce the YAML input, as long as the keywords remain the same, I would need to spend 5 minutes just to prepare the base input, in which the script puts the necessary parameters. For example, when one enters the height of the integration plane, in the base script there is 
```
height[@start@, @out@]
```
And ``sed`` takes care of substituting them with the actual number. 

Adding new keywords would be easy. I have defined a function ``valueInp`` that takes care of the input of a parameter, checks that it is not empty and asks the user to confirm what they entered. For example:

```
printf "\n\nWIDTH OF THE SLICES\n"
delta=0.02
valueInp delta $delta
```

bast:
There is no doubt that they are useful now and that sharing is good.

heikef:
As I can see it the scripts do not interfere with available code. I think it is a good idea to share. Eventually somebody else might update them one day.

PR #133, opened 2018-01-23
rm trailing whitespace

bast:
This goes to the stable branch. Goal of this is to make it easier to compare master and stable branches.


bast:
Don't get alarmed by the failing checks. This branch does not have any automated testing so this does not mean anything here.

heikef:
Ok, then it is probably save to merge.

PR #134, opened 2018-01-25
Several small improvements

mariavd:
Lukas gave some feedback and also found a bug, so I think it is worth the PR.


bast:
What was the bug?

mariavd:
@bast 

In the batch job script I had missed the $wrkdir variable, and it was unable to read the calculation parameters:

```
start=$( sed -n -e 's/^.*in=//p' $wrkdir/calculation.dat | awk '{print $1}') 
```

bast:
OK. As a suggestion, next time I recommend a better commit message. The fact that it is user feedback is secondary and can be in the commit message text but does not have to be the headline. For more serious bugs I recommend to always open an issue and close with a commit so that we can later figure out what problem some commit solved. But no need to redo things here.

mariavd:
Alright, thanks.

PR #136, opened 2018-01-29
Updates for the bash scripts: user-friendliness, separation of their source files from the actual scripts, MF=X option added, gradient calculated separately 

### LINKED ISSUE #135: Bash scripts: user-friendliness, separation of their source files from the actual scripts, MF=X option added
Thank you @heikef  for your comments last Friday. I made some improvements and hopefully now it should be clearer how to use the scripts for your tasks. 

- [x] The directory with scripts is cluttered

I agree and as you suggested, now those base files, from which the actual scripts are generated, are under ``jobscripts/src``. 

- [x] Paths to different versions of GIMIC

I promised to think about your issue where you have many versions of GIMIC with different paths to the binaries. Hardcoding is by far not the solution since after each git pull, it would need to be redone. My suggestion is that you export the path to one of them and call the scripts as they are. If you decide to change the version of GIMIC for a particular run of the scripts, you can export the path to that other binary instead. I think this is reasonable. 

Example:
```
# Assuming that you have three different ones:
/path/to/gimic1/
/path/to/gimic2/
/path/to/gimic3/

# In your .bashrc:
export PATH="/path/to/gimic1/":$PATH
#export PATH="/path/to/gimic2/":$PATH
#export PATH="/path/to/gimic3/":$PATH

$ source ~/.bashrc
$ which gimic
/path/to/gimic1/gimic

# So the script will use this GIMIC 1
# If for the next calculation you want to use GIMIC 2, you comment out the first one and enable the second one:
#export PATH="/path/to/gimic1/":$PATH
export PATH="/path/to/gimic2/":$PATH
#export PATH="/path/to/gimic3/":$PATH

$ source ~/.bashrc
$ which gimic
/path/to/gimic2/gimic
```

Of course, if you insist to use absolute paths in your SLURM job script in the way you have previously used it, it will still work fine, irrespectively of the exported path in your ``.bashrc``. 

- [ ] A copy of the batch job script in the calculation directory 

My intention is that the scripts do the hard work and let the user focus on their calculations, instead of figuring out how bash scripting works. Trying to add fixes without having a good knowledge of it might lead to errors. If something behaves weird, then one should email me or make an issue in the git. Similarly to running DEFINE in Turbomole, you don't bother what exactly it is doing for you. We could discuss further. 

- [x] Ability to restart a prepared but unfinished job

I think this is the solution to the above problem, so I listed it as a separate task for myself. Resubmitting the job to the queuing system with ``sbatch`` requires several extra parameters, coded into the respective bash script. I have worked on restarts, so that if some of the slices are already calculated, there is no need to do them again. It needs some further testing, so this has to wait for now. But yes, it will appear eventually. 

- [ ] Previewing the plane when on a cluster

It is true that X11 forwarding can lag but it changed from unusable to quite decent when I specified compression for the ssh connection. The entry in my ``~/.ssh/config`` looks like this:

```
Host taito
   hostname taito.csc.fi
   User     mariavd
   ForwardX11 yes 
   IdentityFile /home/suzanka/.ssh/deb 
   Compression yes 
   Ciphers blowfish-cbc,arcfour
```

You can try but I cannot, of course, guarantee miracles. How do you usually preview your planes? *xmakemol* is too basic for the purpose in my opinion, unless the situation is obvious, such as in polycyclic planar molecules. At least I have trouble seeing exactly whether the plane unintentionally crosses an atom in complicated molecules. 

- [x] MF = X

Added.

- [x] User-friendliness: some messages were updated. 

- [ ] The documentation was extended but it will come in a separate pull request. For now it can be found on my fork only. 

- [x] The ``GIMIC_HOME`` environmental variable now includes the jobscripts directory, too. Remember to update all of the aliases that you added to your ``.bashrc``.

- [x] Bug fixes:
- Missing characters in the ``anprofile`` function (from the ``useful_aliases`` file) for analysis of the current profile 
- The function which checks whether a value was entered was not returning the checked value properly when there were spaces in between. It was causing a wrong value for the midpoint of atoms in the input of the starting and end points of the integration place.

Let's discuss and then I will make the pull request.

mariavd:
I did not add the (fixes #135), so that we can discuss more.


heikef:
Thank you!

PR #140, opened 2018-03-13
removed edens module closes #118

### LINKED ISSUE #118: edens module not working
We have an edens module, but it is a) not working and b) we have not documentation on how it should work. We should figure out on how this works and decide if we need it or if we prefer to replace it with a new module.

heikef:
The edens module is not working. Nobody replied to the discussion in issue #118 so my impression is that it should be OK to remove it. If needed be we can reintroduce it.


bast:
There is something left - see output from Travis.

heikef:
I saw it and am working on it. For some reason when compiling on my machine I got no complains but Travis found the problem. The reason is that in gimic_interface.f90 the edense module is also used. I removed it.

bast:
It worked for you locally because you had a leftover in the build dir from a previous compilation. It would probably have failed in a fresh build directory.

heikef:
Thanks for the explanation! Good that we have Travis to check. Now all tests are passing. Ready for merge.

PR #145, opened 2018-03-14
Clean-up while working on MPI

bast:
This is work in progress - do not merge yet. This is just so that I get it tested, so that you can track progress, and that so that I can get possibly feedback/suggestions.

### DIFF
--- a/src/fgimic/gimic.F90
+++ b/src/fgimic/gimic.F90
@@ -74,7 +74,6 @@ contains
         call getkw(input, 'magnet', settings%magnet)
         call getkw(input, 'openshell', settings%is_uhf)
         call getkw(input, 'dryrun', settings%dryrun)
-        ! call getkw(input, 'show_axis', settings%show_axis)
         call getkw(input, 'calc', settings%calc)
         call getkw(input, 'xdens', settings%xdens)
         call getkw(input, 'density', settings%density)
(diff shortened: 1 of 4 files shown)


heikef:
OK. So far I see no problem with deleted code. It looks you basically added some more if statements.

bast:
Yes, so far just moving if statements to localize the crashes.

bast:
So after a day of poking in the code I am at a point where we need to plan next steps:
- The code is reading input from stdin but it is not trivial to channel stdin to all workers
- I have tried that only one tasks reads and then settings are distributed but this is hard because of complex data structures which are hard to synchronize

In the long term the data structures need to be simplified.

What could work in the short term is that the input file is not read from stdin but given as a positional argument. This is an easy change but will break all current run scripts :-)

Preferences?

bast:
It's better if we discuss this in the issue, not here - I will move the above question to the issue.

heikef:
Good, let's move to the issue then.

bast:
I will remove here the dead ends and keep what is good and let you know once it's ready for review.

heikef:
OK. Thanks for your work.

bast:
Ready for review. I kept only the cleanup part and the launcher part which can be useful for debugging.

heikef:
Are three commits with "rm dead code" really necessary for the history?

bast:
I will squash them into one.

heikef:
All the remaining changes look fine. Thanks for the discussion. We found some good agreements.

bast:
Cool. If you agree with them, please merge. I will continue work in a new PR.

heikef:
Well, I had to wait for all tests to pass before merging.

Issue #148, opened 2018-03-16
update documentation for open-shell case


heikef:
We have now further visualization available for open shell systems, implying that this should be added to the documentation.

heikef:
The information about cube files could also be removed, since it is not needed any more.

PR #149, opened 2018-03-16
included information about open-shell output into documentation close…

heikef:
…s #148 

1. included in documentation for which programs open-shell calculations are feasible with gimic
2. included information about output files in documentation
3. shortened the output file section 
4. removed information about cube files, since we do not generate them any more


bast:
Awesome!

PR #152, opened 2018-03-21
Feature txt files

heikef:
1. removed txt files from output but kept them for debugging
2. adapted tests and added one new test
3. the new test is added because for practical reasons jmod.txt is written out only for an integration type pre-run when grid(bond) type=gauss and not type=even is applied. 
4. we have now less output files, which is good because for example for 3D calculations the txt files were never used after we moved to VTK type visualization
5. added a warning in the output file when grid type=even is not used VTK files are not printed
6. I think it is reasonable to keep information about the txt files in the documentation since they did not get completely deleted


bast:
Commit message says "removed unnecessary text files" but nothing has been removed or has it?

heikef:
In the test suite the txt files have been removed, but it is still possible to print them out if needed be by switching the debug flag on.

bast:
I think a better commit message would have been "do not print txt format vector fields by default" - are comfortable with changing the commit message?

heikef:
Yes, I agree, another commit message would have been clearer. While working on something you do commits which are clear for you but not necessarily for other. Therefore we have code review. I am fine with changing the commit message. The problem is I do not know how. I know that it is possible to change the latest commit message by using "git commit --amend" but I need to go one commit further back in history.

bast:
First create a backup branch in case something goes wrong. Then edit commits with `git rebase -i origin/master`.

heikef:
Thank you, I will try this out.

heikef:
learned something - replace pick by reword - save and quit and a new editor opens - then change the commit message - save and quit and do a new git push

This is also a useful page: https://help.github.com/articles/changing-a-commit-message/

heikef:
Ready for a new inspection. :)

bast:
Super!

PR #153, opened 2018-03-21
added a check for zero entries in basis information into turbo2gimic.…

### LINKED ISSUE #74: Check of basis set - zero + ordering
We should add a check for the coefficients of employed basis sets. It can happen that a coefficient is set to zero and not to another number like 1 for example. Then gimic gives "NaN" as output because there must be somewhere a division by zero, which is not defined. 

Another issue is that some programs do not have a basis set ordering like s, p, d and so on functions but can have suddenly a polarization function in between. This is known for a while but I do not remember if we take care of this and have a check. Probably not but it would be good to have because gimic assumes s, p, d, f and so on functions ordered by increasing l quantum number. 

The check could be done already on the input conversion level, eg. when calling the turbo2gimic.py script or other scripts. The problems described above are know Turbomole issues related to gimic. 

I know this is not urgent but decided to add the issue to avoid that we forget about this. The information might also be considered useful by some other users.

### LINKED ISSUE #122: QCTools not copied to the installation directory
Turbo2gimic.py needs functions defined in QCTools, which is located in the `tools/` directory. At the moment one needs to copy the folder manually to `bin/` in order to generate the input files from Turbomole calculations. Maybe the setup should handle this.

heikef:
…py and included the QCTools directory into build/bin closes #122 and partly solves issue #74 

1. added a check into the basis section of turbo2gimic.py for checking and removing zero entries, see our discussion at #74 - this does not close #74 because there is no check regarding the spdf and so on ordering for avoiding spdpf ordering. 
2. when working on this and trying to run turbo2gimic.py I ran into the problems described in #122 with QCTools and was playing a bit with the CMakeLists.txt - this works now for me but probably this is a horrible solution for an expert - so please check (it is only one line modified though, so no big harm)
3. albeit being not an optimal solution it is a step in the right direction and could be merged after careful review


bast:
Awesome!

PR #157, opened 2018-03-26
Illustration of the integration plane (fixes #155 )

### LINKED ISSUE #155: Placement of integration plane not clear in documentation
I got feedback from a new gimic user. Apparently the setup of the integration plane for running a gimic calculation is not clear enough. I suggest to improve the documentation. Perhaps a picture is helpful in this context.

mariavd:
What do you think of these pictures? Should I have made one showing both the vectors and the size?


bast:
They look super!

heikef:
The pictures look great! Could you add the fixpoint plus replace start by "in" and include "heigth" and "width" into the picture? Then we would have all keywords included.

mariavd:
Do we prefer to keep the in/out/up/down values positive? At the moment down is defined negative but that is just a leftover from the times when there wasn't ``height`` and ``width``.

heikef:
I think changing in/out values is another issue and not part of this pull request. The figure should not indicate any sign of an input value. It is just a help for the new user to understand what the keywords mean.

mariavd:
Yes, I should I have asked in another topic, sorry.

heikef:
No problem. Often things come up in different discussions, but we need to try to keep things sorted for the sake of keeping an overview.

heikef:
If we keep the "start" label in the figure we should remember to explain that "in" keyword is meant by this.

mariavd:
I'm sorry for that. I used "start" in the script as variable.

heikef:
No need to be sorry, your picture looks fantastic. Would you please remove the obsolete one?

mariavd:
It should be tidier now.

heikef:
Yes, but did you delete the wrong file? I see "start" instead of "in"

mariavd:
Yes, I'm sorry.

heikef:
Perfect! Thanks for your work!

mariavd:
Thanks.

PR #161, opened 2018-04-03
Ports core routines in fgimic to BLAS 1 and 2 (by default off)

### LINKED ISSUE #141: link BLAS and LAPACK to gimic
I think this is not done yet. Please correct me if I am wrong @bast Eventually you have already done this when changing the cmake stuff. Otherwise this is an issue. I would be great to be able to use blas and lapack.

bast:
Fixes #141.

By default BLAS is not used. Added an entry in doc on how to enable BLAS. LAPACK can be enabled similarly but is currently not used. Also added an option for Intel MKL. Depending on the system and the BLAS implementation, the new flag can lead to significant speed-up or slow-down so we need to experiment more.

This is not yet the BLAS 3 rewrite which will be a change of algorithm.


heikef:
Thanks for a lot of work! It is good that we remain flexible in switching the BLAS on and off. As I understand more testing is needed. 
Why did you introduce the uppercase suffix?

PR #162, opened 2018-04-05
Reading large XDENS files

mariavd:
Some time ago Kevin had issues when reading large XDENS files and he recently found a solution for it. He was happy to share it with us, so I suggest that it is included in the master.

### DIFF
--- a/src/libgimic/dens.f90
+++ b/src/libgimic/dens.f90
@@ -80,11 +80,11 @@ contains
             if (ispin == 2) dens=>this%db
             if (this%pdens_p) then
                 do b=0,3
-                    read(XDFD,*) kusse
+                    call rdkusse(XDFD,s*s,kusse)
                     dens(:,:,b)=kusse
                 end do
             else
-                read(XDFD,*) kusse
+                call rdkusse(XDFD,s*s,kusse)
                 dens(:,:,0)=kusse
             end if
         end do
@@ -125,6 +125,14 @@ contains
             end do
         end do
     end subroutine
+
+    subroutine rdkusse(iunit,n,kusse)
+      integer iunit,n,i
+      double precision kusse(n)
+      do i=1,n
+      read (iunit,*) kusse(i)
+      enddo
+    end subroutine rdkusse
 
     subroutine del_dens(this)
         type(dens_t) :: this


heikef:
Thank you, this is a good idea. Could you clean up the two merge commits? This does not look necessary for the history. One commit should be enough. Please update also with gimic master, we have some cmake changes. Sorry for the inconvenience.

heikef:
What was exactly the problem when reading XDENS? Could you explain this plus the solution?

mariavd:
Gimic was crashing on some cluster. I can't remember exactly what was going on but I am about to run calculations with a 4GB XDENS file soon, so if I see the error myself, I will write back. 

I will do the cleanup a bit later (sorry again) because I recompiled with ifort and now I am getting seg faults for all calculations. I am investigating at the moment and if I notice something that needs fixing, I will file an issue.

heikef:
Radovan suggested some changes in XDENS file format. I mention this pull request in the issue discussion. I think this is a bit related.

mariavd:
I'm afraid that I can't fix the branch merging commits. They seem to be on my master branch, not on the one from which I am making a pull request. If you don't like them, just ignore the request.

bast:
We can squash the commits here so this is not a problem. The question is rather, why does your master branch diverge "This branch is 2 commits ahead, 18 commits behind qmcurrents:master." You commit to your master branch and that is not a good practice :-)

mariavd:
I made the new branch from master but I forgot to change to it, so I ended up committing to master. I used hard reset and in the end I again ended up these merge commits.

bast:
The repair is the same as last time: create backup branch, fix your master branch, rebase the feature branch. But it's OK for me to squash here if you don't want to go through that.

heikef:
What is the status here?

mariavd:
If the modification is accepted, I can rebase.

bast:
Please remove the out-commented code and what does "kusse" mean?

bast:
Also `!kr` should not be there. But I understand that you are the messenger so I don't want to give you additional work - I can also clean this up after the merge is in.

bast:
Now I remember ... this is what we were debugging together in Espoo last year.

bast:
And we arrived at same conclusion AFAIR. So this is a good change, just needs cleanup but don't be afraid of delegating the cleanup.

heikef:
If @bast has been involved in this before I leave the final review to him. Reading that this has been worked on since last year my impression is that it is time to share. :)

mariavd:
Apologies for the late reply. Yes, this is the routine which reads the XDENS file. I don't understand the cryptic name kusse - it is how it existed originally in gimic. I'm taking care of the cleanup.

bast:
Thank you. Can you please also rename it to something more meaningful? But also OK to not do this and then we do this in a separate step. I don't want you to suffer for "bad" choices of others :-)

mariavd:
No problem. I'll fix it during my lunch break.

bast:
Lunch break is for lunch. I will merge this and fix this in a separate step. Thank you!

Issue #165, opened 2018-04-12
Segmentation fault with ifort


mariavd:
Last week I noticed that if I compile gimic with ifort current density calculations crash with a seg fault but bond integration works fine. So, I compiled with traceback to find out what's going on, since with gcc the calculations finish. 

```
./setup --fc=ifort --extra-fc-flags='-traceback -check all' --cc=icc --cxx=icc
```

I noticed two things. 

1) The ``rank`` variable in the file below is not initialized but used in a subroutine. Probably it needs to be defined as intent in. 

```
forrtl: severe (194): Run-Time Check Failure. The variable 'initialize$RANK$_114' is being used in '/homeappl/home/mariavd/appl_taito/gimic/src/fgimic/gimic.F90(96,13)' without being defined
Image              PC                Routine            Line        Source
gimic.bin          00000000004063E8  gimic_IP_initiali          96  gimic.F90
gimic.bin          000000000040460A  MAIN__                     44  gimic.F90
gimic.bin          000000000040455E  Unknown               Unknown  Unknown
libc-2.12.so       0000000005C4ED1D  __libc_start_main     Unknown  Unknown
gimic.bin          00000000004043E9  Unknown               Unknown  Unknown
```

2) In the jvector_plots subroutine the pointer to the tensor jtens is used but not allocated. The function get_acid(r, tens) takes the pointer to the tensor as intent(in), so it receives an undefined pointer. This problematic subroutine is called by the run_cdens subroutine. I actually don't know how the code can work when compiled with gcc with this issue. 

```
forrtl: severe (408): fort: (7): Attempt to use pointer JTENS when it is not associated with a target

Image              PC                Routine            Line        Source
gimic.bin          00000000004BDB60  Unknown               Unknown  Unknown
gimic.bin          000000000045466B  jfield_class_mp_j         326  jfield.f90
gimic.bin          0000000000405689  gimic_IP_run_cden         203  gimic.F90
gimic.bin          0000000000405B5E  gimic_IP_driver_          178  gimic.F90
gimic.bin          000000000040460F  MAIN__                     45  gimic.F90
gimic.bin          000000000040455E  Unknown               Unknown  Unknown
libc-2.12.so       00007FC3D9DC2D1D  __libc_start_main     Unknown  Unknown
gimic.bin          00000000004043E9  Unknown               Unknown  Unknown
```

mariavd:
[...]

PR #166, opened 2018-04-13
Introducing tests with Valgrind - fixes #113

### LINKED ISSUE #113: Add Valgrind test
@mariavd volunteered to take care of this - an important issue to monitor performance

mariavd:
I added the tests and now it's time to test them online with Travis, too. Locally the bond integral test passes successfully. The 2D current density test fails because as I said in the other issue, there is the uninitialized values problem. I will take a look into that after this testing is sorted out. 

I will rebase after everything looks good, so for the moment you can ignore the pull request. 

@bast Runtest is indeed a very useful tool.


test/CMakeLists.txt
```diff
@@ -1,25 +1,5 @@
...
-    benzene/int-grid-bond-even
-    benzene/int-cdens
-    benzene/diamag-off
-    benzene/paramag-off
-    benzene/giao-test   
-    c4h4/integration
-    c4h4/read-grid
-    open-shell/3d
-    open-shell/integration
+    benzene/valgrind-integral
```
bast: But here all the other tests disappeared.

test/benzene/valgrind-integral/test
```diff
@@ -0,0 +1,50 @@
...
+#     get_filter(from_string='definitely lost',
+#                num_lines=1,
+#                rel_tolerance=0),
+#     get_filter(from_string='indirectly lost',
+#                num_lines=1,
+#                rel_tolerance=0),
+#     get_filter(from_string='possibly lost',
+#                num_lines=1,
+#                rel_tolerance=0)
+#     ]
```
bast: Please remove the out-commented leftovers.

test/CMakeLists.txt
mariavd: Yes, because I don't know yet how to split the tests into two parts, so that the other tests are executed with the release binary and those with the debug binary. I was thinking of something like  exporting TRAVIS_TEST=2 to differentiate between the two.

test/benzene/valgrind-integral/test
mariavd: Do we want to also check the value of the integrated current in the debug tests or just the memory leak report?

test/CMakeLists.txt
bast: OK but if this is work in progress then please add "WIP" to your title or make it clear somewhere. Otherwise this could have gotten accidentally merged. I would probably add a CMake option to differentiate.

test/benzene/valgrind-integral/test
bast: Also the correctness.

mariavd:
@bast I think the tests are functioning fine now. They fail because of the undefined variables that exists in the code. They are only detected in the debug release. My next task will be to sort them out. I think the pull request can still remain open until everything gets clean.

.travis.yml
```diff
@@ -12,22 +12,28 @@ matrix:
...
 
 script:
-  - ./setup
+  - ./setup 
   - cd build
   - make -j2
   - make install
-  - ctest -j2
+  - ctest -j2 -E valgrind 
+  - ./setup --type=debug build-debug
```
bast: I think we should not run both release and debug inside one but rather run a matrix of two build.

test/benzene/valgrind-integral/test
```diff
@@ -0,0 +1,38 @@
...
+     get_filter(from_string=' *** Integrating current',
+               num_lines=10,
+               rel_tolerance=1.0e-8),
+    ]
+
+ierr = run(options,
+           configure,
+           input_files=['gimic.inp'],
+           filters={'valgrid-summary.dat': f,
+	            'stdout': acc})
```
bast, line 34: I would probably not check the Valgrind output through the entire runtest infrastructure but here write a short custom Python function which only checks that there are no errors and not leaks.

mariavd: I'm not good at Python, so maybe I'm missing something. The output from Valgrind needs to be parsed in some way or another. If I had to do it in a separate script, then wouldn't it have to replicate what runtest does?

.travis.yml
mariavd: So you mean that both the debug build and the release build sould run all tests?

test/benzene/valgrind-integral/test
bast: OK - I don't want to complicate it. Finish what you had in mind, later we can always adjust if really needed but it's probably better as you have done.

.travis.yml
bast: Not necessarily, but they should not run inside the same test but be two independent tests from the Travis point of view. If this is not clear, then don't worry about it - I can always split this later.

mariavd:
@bast I think I managed to make two separate tests for the debug and release builds. Debug fails due to the seg faults, and release passes as it used to. Is this what you had in mind? I can change it if I misunderstood.

bast:
OK can you please clean up the history before we merge?

mariavd:
Done. Merging with master will mean that tests are reported as failed until I sort out the uninitialized variables and the j tensor pointer. I will continue later. I think the PR should remain open for now.

bast:
Yes about the red tests if we merge. We can wait but also the cleanup should not be only on you. Unless you prefer to work on it. Please let me know how we should proceed.

bast:
A minor comment on commit message: convention is one line no longer than 50 characters, then if more explanation needed, one blank line followed but free-form text.

mariavd:
Thanks for the note. I updated the commit message.

We can share the work on the problems, I don't insist on fixing everything myself but I will be glad to give a hand.

bast:
OK. Please let me know when I should take over. Alternatively open issues for each problem, self-assign those that you wish to work on and then the rest is up for grabs.

mariavd:
OK, I'll start opening separate issues.

bast:
Should we merge it even though it makes tests fail? Since they fail for good reasons.

heikef:
I am not sure if this is a good idea.

bast:
So you prefer if we don't?

heikef:
A new user might think that nothing works.

bast:
So - what do you prefer?

heikef:
Would it be possible to fix it in this pull request? Or create several pull requests that pass all tests?

bast:
Yes it's possible. Alternative: we create a branch, merge the PR into that branch and then fix stuff on that branch.

heikef:
Yes, creating an extra branch for this is a good solution. When we know that things are working we merge this branch into master. Then we avoid failing tests and half finished things on master.

mariavd:
I suggest that maybe sorting out the problems in issue #165 is the first step to take, so this request can be left open. I can take a look tomorrow and comment on those issues.

bast:
Hopefully this will become green once #167 is in.

bast:
Now I recommend to rebase this on top of master.

mariavd:
I merge-rebased master and the log looks normal. Do I need to add some commit to make Travis re-run the test?

bast:
No, just push the rebased branch and Travis will restart since the branch changed.

mariavd:
Sorry, I'm again doing something stupid. I did 
```
git fetch upstream
git merge-base upstream/master origin/master
```
The log looks fine, I think:
```
*   871267d - (22 minutes ago) Merge pull request #167 from bast/radovan/warnings - Radovan Bast (upstream/master)
|\
| * af0d84f - (2 hours ago) use mpi_rank instead of rank - Radovan Bast
| * e467136 - (3 hours ago) do not use uninitialized jtens - Radovan Bast
| * ce2b30d - (3 hours ago) dead code has to go - Radovan Bast
| * 0ec6251 - (3 hours ago) get_acid does not use r, remove it - Radovan Bast
| * a0cd8b5 - (3 hours ago) workaround for compiler warning - Radovan Bast
| * 683cb21 - (3 hours ago) avoid compiler warning about uninitialized prsp1 - Radovan Bast
| * 96c7db3 - (3 hours ago) remove unused r parameter - Radovan Bast
| * e50700a - (3 hours ago) remove unused mol parameter - Radovan Bast
| * c24db4a - (3 hours ago) remove unused code - Radovan Bast
| * 2abd3b1 - (3 hours ago) avoid tabs - Radovan Bast
| * 4c5c588 - (3 hours ago) avoid comparison of real number and zero - Radovan Bast
| * d51c943 - (3 hours ago) mol unused in setup_std_grid - Radovan Bast
| * 7e00784 - (3 hours ago) replace tab with spaces - Radovan Bast
| * 4bec19d - (3 hours ago) nullify grid pointer in del_integral - Radovan Bast
| * 8df5448 - (3 hours ago) fix setting of is_mpirun in init_mpi - Radovan Bast
|/
| * ec5fc3e - (5 days ago) Prepared valgrind test for bond integration and for 2D current density on benzene. - mariavd (origin/valgrind, valgrin
|/
*   d5b1ec4 - (6 days ago) Merge pull request #164 from bast/radovan/cleanup - Radovan Bast (HEAD, origin/master, origin/HEAD, master)
|\
| * 8abbdaf - (6 days ago) rename cryptic "kusse" to "array" - Radovan Bast
|/
```

But then when I git push, it says that everything is up to date. But on my fork on Git it says that it is behind 16 commits.

bast:
```
git checkout master
git pull upstream master
git push origin master
```

mariavd:
Thanks!

bast:
`merge-base` does something very different: https://git-scm.com/docs/git-merge-base

mariavd:
Thanks again, I will continue sorting out the issues later today or tomorrow. I think I can handle it.

bast:
Now this PR contains my commits. What precisely did you do to rebase?

mariavd:
I though that 
```
git rebase master valgrind
```
is what will move the valgrind branch to point to the latest commit on master.

bast:
I think you rebased correctly, then you tried to push, then it did not work, then you pulled and that led to your local valgrind branch diverging from remote valgrind branch.

When you rebase you change history, then you cannot push, you have to force push. Of course that is a dangerous operation so always create backup branches.

mariavd:
Yes, at one point I couldn't push and I thought I shouldn't force push, then I was googling for some seemingly not useful advice and eventually I could push without force... I'm sorry about the mess. Maybe the easiest solution is to copy the files elsewhere, and then reset the branch before yours and commit?

bast:
Continue work, once you are ready with this PR do a `git reset --soft master`, then commit all your changes in a nice commit.

mariavd:
The tests pass now after the commit with the change in jfield.f90. Maybe it should not be squashed to the previous since it is a bug fix, not related to the actual test. 

One final question, where are the FC flags for the debug release defined?

heikef:
When integrating over a bond one typically does not need the "radius" keyword. It has to be used in the input file and commented in. This option is helpful when dealing with very dense currents, when for example a rectangular integration plane would also cover neighboring currents. Long time ago we were also discussing the option to allow an ellipse instead of a circle.

bast:
`git grep` for current flags to see where they are set. But they should not be changed there. We should introduce a new file for this. I recommend you open a new issue with flags you would like to have and we close this one. Then I can show how to do that in a PR.

heikef:
radius is part of the grid_t type construct - there is no grid "mode radius" - grids can be either "even" or "gaussian"

mariavd:
@heikef I did not change any functionality. In grid.f90 ``this%grid%mode`` keeps what kind of grid is selected. Please check line 74. If one does current density calculation in 2D or 3D one uses the standard grid mode, where the basis vectors are entered by the user in the input file. When we do bond integration we use a bond grid. In short, I mean that grid mode is different from grid type. 

The radius variable is defined in the grid_t structure but it is only initialized when one uses bond integration. The extra line that I added simply makes sure that the uninitialized radius variable is not compared to something, since this is a dangerous operation. Do I need to explain further?

@bast Thanks, I will ask about the flags separately.

heikef:
My point is that "radius" was never intended for something else than investigating the currents over a bond. It does not make sense otherwise. The reason why "even" grids have never been considered is that it does not make sense to use even grids when gaussian grids are available because the gaussian ones lead to much more accurate results.

mariavd:
I'm confused. Should I do something about it? I did not mention gaussian vs even grids but bond vs standard grids.

heikef:
@mariavd I am able to read the code and I see the changes done to it. I thought explaining on why the radius stuff was set up as it is might be helpful. It was only intended for helping at the bond integration and there it worked fine. Thus, calling your extensions a bug fix is debatable in my view.

mariavd:
Right, thanks. Is there anything else to do regarding this PR?

bast:
From my perspective this is ready to go. Thanks for all the work!

mariavd:
@bast  Thank you very much for the help!

PR #171, opened 2018-04-18
restore OpenMP parallelization; closes #168

### LINKED ISSUE #168: Restore OpenMP parallelization
Currently all tests pass using OpenMP when compiled without optimization.

With optimization the following test fails:
- benzene/integration

bast:
The important change is `OMP PARALLEL DEFAULT(NONE)` which is very important to do.

### DIFF
--- a/src/fgimic/integral.f90
+++ b/src/fgimic/integral.f90
@@ -103,17 +103,19 @@ contains
         xsum3=0.d0
         psum3=0.d0
         nsum3=0.d0
-!$OMP PARALLEL PRIVATE(i,j,k,r,rr,xsum,psum,nsum) &
+
+!$OMP PARALLEL DEFAULT(NONE) &
+!$OMP PRIVATE(i,j,k,r,rr,xsum,psum,nsum) &
 !$OMP PRIVATE(jt,w,jp,tt,jvec) &
-!$OMP SHARED(xsum2,psum2,nsum2) &
-!$OMP SHARED(p1,p2,p3,this,center,spin,bb,normal,mol,xdens,lo,hi) &
-!$OMP REDUCTION(+:xsum3,psum3,nsum3)
+!$OMP SHARED(p1,p2,p3,this,center,spin,bb,normal,mol,xdens,lo,hi,bound) &
+!$OMP REDUCTION(+:xsum3,psum3,nsum3,xsum2,psum2,nsum2)
         call new_jtensor(jt, mol, xdens)
         do k=1,p3
             xsum2=0.d0
             psum2=0.d0
             nsum2=0.d0
-            !$OMP DO SCHEDULE(STATIC) REDUCTION(+:xsum2,psum2,nsum2)
+
+            !$OMP DO
             do j=lo,hi
                 xsum=0.d0
                 psum=0.d0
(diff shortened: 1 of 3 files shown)


heikef:
Could you explain why this is important? How does one decide what is openmp shared reduction and private? I consider this as an opportunity for learning If answering requests too much of  your time I am happy with a link on where to read more on how a code is modified for openMP. Thanks in advance.

bast:
With `default(none)` you have to specify every variable explicitly and there are no defaults. This is more typing work but much safer since you cannot forget any variable.

bast:
About shared/private: that's a longer story. If variables a private, you cannot have race conditions. But you don't want to make everything private since then memory is duplicated. It is important that you don't write to shared variables unless you mark it as a reduction.

heikef:
I see mol and xdens are also declared, but they are not used variables inside the loop. Why do you have to include them in the declaration?

bast:
Can you comment directly on the line of the code? Then I know what you mean precisely.

bast:
"Files changed" -> comment there.

heikef:
integral.f90 line 110 for example --> ok commented in "files changed" directly

src/fgimic/integral.f90
```diff
@@ -103,17 +103,19 @@ subroutine integrate_current(this, mol, xdens, spinn)
...
         nsum3=0.d0
-!$OMP PARALLEL PRIVATE(i,j,k,r,rr,xsum,psum,nsum) &
+
+!$OMP PARALLEL DEFAULT(NONE) &
+!$OMP PRIVATE(i,j,k,r,rr,xsum,psum,nsum) &
 !$OMP PRIVATE(jt,w,jp,tt,jvec) &
-!$OMP SHARED(xsum2,psum2,nsum2) &
-!$OMP SHARED(p1,p2,p3,this,center,spin,bb,normal,mol,xdens,lo,hi) &
-!$OMP REDUCTION(+:xsum3,psum3,nsum3)
+!$OMP SHARED(p1,p2,p3,this,center,spin,bb,normal,mol,xdens,lo,hi,bound) &
```
heikef, line 110: here mol, and xdens are defined but they are not used in the loop - why do you have to do this?

bast: Much better - thank you ...

bast: All variables inside the parallel region need to be defined.

heikef: I know this has also been done before but I was wondering why.

bast: Threads spawn in the parallel region but the actual concurrent work is done inside the DO-loop.

heikef: I thought everything in the "parallel region" is inside the loop. This is apparently wrong. How do you decide on what needs to be in the parallel region?

bast: The parallel region could be just the loop. But creating and closing parallel regions costs time so you don't want to create them too often either. As small as possible, as large as necessary.

heikef: So you define it at the beginning of the routine and at the end only to save time. 

thread spawning = creating a new thread
thread =  smallest sequence of programmed instructions that can be managed independently by a scheduler

heikef:
I learned something. Thank you!

PR #178, opened 2018-07-12
ETA calculated in hours and days

mariavd:
Print the ETA in days if it will be more than 48 h

Expanded the fields for GTO and wall clocks in hours and days


mariavd:
Actually the ETA can also be divided by the number of requested threads. 
@bast I could not find the variable that stores the OMP_NUM_THREADS after a fairly quick look. Could you maybe tell me what its name is?

mariavd:
Sorry, I forgot to commit some changes. That's why the tests fail.

bast:
Thank you!

PR #183, opened 2018-07-20
Interactive scripts: vertical profile, squares profile, restarting (fixes #158)

### LINKED ISSUE #158: Bug in the scripts
I found a bug in the current profile script, so it should be sorted out before somebody ends up with issues. I can modify them in a bit of a stupid way or the other option is to modify the output message for the modulus of the current. At the moment it looks like this:

```
 INFO: Closed-shell calculation

 Integrating current density
 *****************************************
 *** Integrating |J|

 ************************************************************
 Induced mod current (au)   :     0.418819
       Positive contribution:     0.595170  (  16.771552 )
       Negative contribution:    -0.176352  (  -4.969481 )

 Induced mod current (nA/T) :    11.802071
       (conversion factor)  :    28.179409
 ************************************************************


 *** Integrating current

 ************************************************************
    Induced current (au)    :     0.418821
       Positive contribution:     0.595166  (  16.771417 )
       Negative contribution:    -0.176344  (  -4.969283 )

    Induced current (nA/T)  :    11.802134
       (conversion factor)  :    28.179409
 ************************************************************
```

Do you mind if I change it to "Positive mod contribution" and "Negative mod contribution"? 

The problem is that there are two instances of "Positive contribution" in the output file - both for the J and its modulus and when I use grep, it finds both.

mariavd:
Here are updates and bug fixes to the scripts.

Major additions in short: 
- There is a script to make a current profile vertically with respect to two given atoms. 
- There is a script to split the integration plane into squares. It works in some limited cases, so it needs further work. 
- Current profiles can be restarted on clusters without losing your partially completed calculations.

(I had complained about a problem with the integral.f90 file but it seems that it is no longer present.)


mariavd:
I resolved the conflicts and it seems that I need to click on ``commit merge``, meaning that it might accept the changes. Is that fine?

mariavd:
I hope I didn't mess up with the above merge. I thought that since I know my work, I would be able to fix it properly but I think I made things worse. I'm sorry. How can I fix it?

heikef:
At the moment it looks as if there are no problems. Why are you concerned?

heikef:
The X option was not accepted during my test. I am running it now without.

heikef:
This was working. I will merge.

mariavd:
@heikef  So is MF=X working? And I've asked before, why bother with such an unclear and poorly defined keyword? I included it on request but I think new users should not be advised to use it.

bast:
What does it mean? :-)

heikef:
The X option automatically sets the "optimal" B field direction. I use it all the time and so do other user. I do not see why new users should not use it. It is just a matter of taste.

bast:
What does "MF" mean?

mariavd:
@bast Exactly this is the problem in my opinion. This X option is confusing with its definition X, to begin with (MF = magnetic field). It doesn't mean align to the X axis but... gimic does some mystery to find the best direction which is not documented. What happens in twisted molecules? This is the reason why I have been in touch with Lukas and Raul - their programs ask for the user's input as to what atoms the B field should be optimized with.

bast:
I agree. Not a good keyword naming.

heikef:
The X option sets the B field automatically perpendicular to the bond you look at. I can find the place in the code where this is done. A lot of things are or better used to be not documented in gimic. The X option used  to be called "T" option in the very old gimic. At one point Jonas changed it to X. I do not remember why he chose X but I remember that I preferred T. The T or X option is very convenient in particular when looking at tilted molecules.

heikef:
The alternatives of Raul and Lukas might eventually be better but this is no reason to throw the X option out of gimic. As it is none of the options of Raul and Lukas are part of gimic.

PR #184, opened 2018-07-23
added keyword to switch off the jmod integration closes #182

### LINKED ISSUE #182: Reintroduce option to switch off modulus integration
Add mod=off option - typically the current density integration is good enough. Modulus integration used to be helpful for planar molecules to test that the integration works properly. Allowing to switch it off will save time.

heikef:
The keyword jmod=off switches the jmod integration off. I added this keyword into the section  "essential", because this keyword has nothing to do with the section "grid" where it could also have been added. A test example is included in the test suite. It was necessary to adapt the test example for the integration for open shell systems, because doing a jmod integration is not our default any more.


bast (approved):
Looks good to me. Thank you!

heikef:
You could perhaps merge it? :)

PR #186, opened 2018-07-25
corrected typo in usage.rst

heikef:
The header "running gaussian" had a typo. We had Gausian instead of Gaussian there.


heikef:
@bast I think here is not a review needed but since we always use this procedure I assigned you. Sorry in case I wasted your time.

bast:
Peer review is always good. Merging ...

heikef:
Thanks!

PR #187, opened 2018-07-26
added old scan scripts

heikef:
I am doing more and more user support related to the interactive jobscripts we share on the github. Therefore, I decided to share some old bash scripts explaining on how we used to do integration scans with gimic. They are not meant for usage but have a more didactic purpose to help new users to understand what we mean with "integration scan". I think they are self explaining.


bast:
Thanks!

PR #189, opened 2018-07-30
deleted unused code in jfield.f90

heikef:
While trying to localize a bug I realized that we have some unused code in jfield.f90 and decided to clean it up. It is just in the way when trying to spot other things.


bast:
I like that :-)

PR #194, opened 2018-09-25
use runtest v2.2.0

bast:
`runtest` received a couple of bugfixes and a new feature: https://github.com/bast/runtest/releases/tag/v2.2.0


bast:
Other bonus of this change is that we fetch a well-defined version from PyPI, instead of a moving version from GitHub.

heikef:
Thank you!

PR #198, opened 2018-10-09
trivial: Rehuel Lobatto with a double-t

lnw:


heikef:
Thanks for the typo fix.

PR #205, opened 2018-10-18
do not fail configuration if not inside a git repo; fixes #204

### LINKED ISSUE #204: configuration fails if git is present but the directory is not under version control
During the cmake configuration, there is a test if a git binary exists, and if so, the version of the current commit is retrieved.

However, it is perfectly possible to obtain the zipped project from github, so it is not under version control.  If during the configuration a git binary is found, the configuration fails.

bast:


heikef:
Thanks for fixing this issue!

PR #206, opened 2018-10-23
Integration plane parameters can be changed without restarting

mariavd:
I have done a lot of new additions for the scripts, removed some bugs here and there and most importantly, now it is not necessary to restart the whole script if one wants to modify a parameter, e.g. after previewing the integration plane.

### DIFF
--- a/jobscripts/src/current-profile-header
+++ b/jobscripts/src/current-profile-header
@@ -3,6 +3,33 @@
 # Function definitions:
 source  @SCRIPTS_DIR@/src/functions-def
 
+function createInpTemplate() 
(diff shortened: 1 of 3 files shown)


heikef:
Thanks for the work, Maria! jmod=off/on should not be in the "Advanced" section, it is part of "Essential". Could you please remove the comment?

mariavd:
Moved. Good that you noticed. It should go to the documentation. I have edited a couple of small things there but I have more in mind, so it will eventually come.

heikef:
Ok, thank you.

Issue #209, opened 2018-11-28
Add macOS to the automated testing


heikef:
It would be nice for Mac users if gimic could be operating on a Mac.

bast:
We should add a macOS test on Travis for this.

heikef:
During a new attempt to install gimic on my Mac I ended up doing a cleanup of unused code. Afterwards I was able to compile. The problem with the Mac is the library mess.

heikef:
The following was helpful in my case: 
1. use Anaconda and install everything needed in requirements.txt
2. gfortran, gcc, g++ have to be in /usr/local/bin/
3. in the .bashrc I had to add:
 # for the compilers:
 export PATH="/usr/local/bin:/usr/bin:/bin:/usr/sbin:/sbin"
 # for Anaconda: 
 export PATH=/Users/my_username:/Users/my_username/anaconda/bin:$PATH
 # for CMake:
 PATH="/Applications/CMake.app/Contents/bin":"$PATH"

heikef:
I tried several things using MacPorts and Brew and XCode. The main problem was to set the PATH correct such that gimic cmake finds what it should. I am sure there are better solutions possible compared to the one which worked for me.

heikef:
The setup --omp stuff was also working, but not the --blas option. This is because it does not find the Blas and Lapack libraries. Any idea on how to install them on a Mac + make it work for gimic?

heikef:
By the way I used Anaconda 2.

heikef:
I was thinking about this issue. Is it really a good idea to support different operating systems? Should we not instead concentrate on improving gimic?

bast:
But users install GIMIC also on macOS so the automated testing will hopefully help us to make GIMIC more robust also for macOS.

heikef:
Does this imply we include automated testing for Mac on TravisCI but this will not affect other non-Mac users when installing the code on their local machines?

bast:
Right.

heikef:
Cool, but how do we do this?

PR #212, opened 2019-01-12
removed div_j and edens from interface

heikef:
This is a cleanup of unused stuff in the interface files. I tried to make gimic work on a Mac and during this procedure I was doing a cleanup. Afterwards gimic compiled on my Mac and all tests except the usual Valgrind suspects ran through.


heikef:
This does not close #209 but contributes to it in my view.

bast:
Thank you for the cleanup!

PR #213, opened 2019-01-27
added python scan scripts

heikef:
After a discussion with @lnw I decided to share some python scripts to do command line gimic scans. For me this was a python coding exercise but maybe it is a good idea to share them to improve their quality and to learn something. As they are they are not very nice.


bast:
Thanks! I will merge them now. There is always room for improvement but this can then be done in subsequent PRs.

PR #214, opened 2019-02-13
Scripts bug fixes

mariavd:
Several problems are fixed that prevented smooth use of the scripts

Error in the extraction of the critical pts in the profile
Removed some notification messages when setting up the current profile
The SLURM script ensures that each single task is run on one core


heikef:
Thank you Maria! Would it also be possible to add a test for your scripts? I understand if this is not quickly possible for the present pull request but you should keep it on the radar.

mariavd:
Yes, this is a good idea. I will prepare a test after I am done with my thesis.

heikef:
OK. Thank you for your work.

Issue #170, opened 2018-04-18
Timing summary is incorrect for an OpenMP calculation


bast:
For an OpenMP calculation we should use OpenMP timers.

heikef:
Did you think of ''omp_get_wtime()'' ?

bast:
Yes exactly that.

heikef:
https://gcc.gnu.org/onlinedocs/libgomp/omp_005fget_005fwtime.html

heikef:
For the test cases?

bast:
In general, does not matter for test cases. The timing output for any GIMIC calculation is wrong if run on OMP.

bast:
But it's not a problem for me now.

heikef:
Ok. Thanks.

PR #216, opened 2019-03-19
fix estimated time

lnw:
1) fix wrong ETA, and make obvious why there is a factor
2) replace multiple divisions by multiplications (so it is obvious that the order of execution is irrelevant)


bast (approved):
Thank you!

heikef:
Is this solving as well issue #170 ? Thanks for you effort, Lukas!

lnw:
No, that is unrelated. #170 is about wrong final times in case of OMP, but this patch just fixes the estimated time that is printed at the beginning of a run.

heikef:
Ok. Thank you.

PR #221, opened 2019-03-29
Feature update docu

heikef:
Updated the installation section plus visualization section after receiving user feedback.

### DIFF
--- a/doc/installation.rst
+++ b/doc/installation.rst
@@ -3,6 +3,24 @@
 Installation
 ============
 
+Before compiling GIMIC you need to make sure that you have installed
+ideally all of the packages collected in
+the ``requirements.txt`` file.
+You need minimum the ones listed below. ::
+
+* cython
+* numpy
+* runtest == 2.2.0
+
+A convenient way to install the packages listed in ``requirements.txt``
+is to install the 
+Anaconda2 (https://www.anaconda.com/distribution/) Python distribution
+first. 
+Then you can simply install all the packages listed in the file
+``requirements.txt`` by using:: 
+
+  $ conda install name-of-package
+
 GIMIC requires CMake to configure and build. CMake is invoked via a front-end script called ``setup``::
 
   $ ./setup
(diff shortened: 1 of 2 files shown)


doc/installation.rst
```diff
@@ -3,6 +3,18 @@
...
+Before compiling GIMIC you need to make sure that you have installed
+all of the following packages, see also the ``requirement.txt`` file.
+You need at least cython and runtest. ::
+
+* cython
+* numpy
+* pyparsing
+* pyyaml
+* sphinx
+* sphinx_rtd_theme
```
bast: pyparsing, pyyaml, sphinx, and sphinx_rtd_theme should not be listed here. I will at some point switch to pipenv where it is clearer which packages are for running, which for testing, and which for developing.

heikef: I included: You need at least cython and runtest -- For the time being this looks ok to me. We can always update the documentation another time.

bast: "Inexperienced" users will however try to install these with sudo. I think we should really remove these from this list.

heikef: Maybe we should recommend to install Anaconda2 and use conda install XXX to get all the packages. Inexperienced and pragmatic users would like this.

heikef: Just to be clear: I do not mind removing the stuff from the list. Instead I try to figure out a user friendly installation way.

bast: I agree.

heikef: Would it be acceptable to recommend installation of Anaconda2 then? If yes I will update the documentation and file a new pull request.

bast: Yes I think this is a good idea. Better than letting users sudo install things.

bast:
One more nit-pick: it's `requirements.txt` instead of `requirement.txt`.

heikef:
Sorry. I will fix this.

bast:
Thumbs up! Do you want me to squash-merge?

heikef:
No, I will do it myself.

bast:
Super - thank you!

PR #223, opened 2019-03-30
added Mac installation guide

heikef:
I am sharing here my experience with installing Gimic on my Mac. At the end it worked out but I have no clue why and how. :) However, since I saw on the issue list that people face similar problems I thought it is a good idea to share. Perhaps there is a Mac user out there who is able to figure out a much simpler and better way to get it work.


bast (approved):
Thanks for sharing!

PR #226, opened 2019-03-31
Valgrid tests are optional

bast:
On Travis CI they are run but by default they are off.


heikef:
Great! Thank you. Is it picky to suggest to squash everything into one commit?

bast:
I think in this case the commits do different things so I think it is better not to squash them in this case. Only the last commit is really about the Valgrind, the rest is cleanup.

heikef:
Do you really think that commits like "rm unused code" and "rm more unused code" are needed?

bast:
For sure they are unrelated with the Valgrind commit. But it is totally OK for me if you prefer to squash them, then go ahead.

heikef:
I am not able to do this online. I could accept the pull request from you and then squash the commits and then push again but without requesting code review. Would this be acceptable?

bast:
Why can't you do this online? (in the "merge pull request" you have the choices)

But still I would here recommend to not squash, but merge as they are.

heikef:
Ah, this is too tempting...

heikef:
I was not aware that this option exists.... what a feeling! :) It is much easier to do this online as a reviewer....

PR #228, opened 2019-04-04
adapted getkw.py and gimic.py.in to python3 closes #54

### LINKED ISSUE #54: Support for Python 3
Compiling gimic on abel is not straightforward, but possible.  The basic problem is, that gimic is really limited to python2, but on abel some python packages only exist as python3 (eg. numpy).

The following works for me:

- load modules:
    - python 3.5
    - cmake
    - intel

- set some aliases:
    - python=python3
    - PYTHON_INCLUDE_DIRS=/cluster/software/.../...3.5m

- setup --fc ifort --cc icc --cxx ccpc mybuild

- cd mybuild; make; make install

- manually change the shebang of all python scripts in mybuild/bin from 3.5 (which was set automatically based on the previous choice) to 2 (which they really are).

It would simplify things a lot, if gimic supported python3.  That refers to cmake as well as the actual scripts.

heikef:
Adapted gimic to python 3. Note, this might perhaps create problems for people using python 2.


lnw:
Did you find out why that extra treatment of `nargs' is required?  If there is a legitimate reason for getting args which is neither of int, list, tuple, None, shouldn't that reset to -1 go to the end of __init__?

heikef:
@lnw That is exactly what I did. When nargs is none then it is set to -1.

src/fgimic/getkw.py
```diff
@@ -316,10 +315,13 @@ def setkw(self, arg):
 			self.arg=[]
 			return
 
+		# try to avoid a none value for self.nargs
```
heikef, line 318: @lnw here

lnw:
I saw those lines, that's what I'm asking about:  Is this really where the problem comes from or is this a symptom of another problem? I can see that your fix removes the error you are getting, but I'm not sure it's the right solution.  More specifically, what I don't understand is: how can args be neither of int, list, tuple, None?  Why wasn't that a problem before?

heikef:
Thanks for that input point. In fact I can not answer this. I was also suspicious in this respect. Therefore, I made sure that the test suite is running through without problems.

tools/python-scripts/get_input_py3.py
```diff
@@ -0,0 +1,251 @@
+#!/Users/heike/anaconda/bin/python
```
lnw: this might not work for most people ...

lnw: ;)

heikef: Oh no! :)

heikef: Adapted. Thanks for the feedback Lukas!

heikef:
Ok all tests passed and in a sense I got a code review by @lnw - would it be ok to merge @bast or would you prefer to join the discussion? Since there was an open question we were not able to solve I hesitate to do the merge myself.

tools/python-scripts/get_input_py3.py
lnw: /bin/python also doesn't work everywhere.  Default debian for example is /usr/bin/python and even then one might want to use a different python.  The most robust solution I know of is '/usr/bin/env python'  but I don't know if that works for example on osx.

I didn't read all of your code, someone who knows more than me should do that ...

heikef: I think somebody using the scripts should know enough python to adapt them. Thanks for the suggestions but I prefer not to change the first line again.

heikef: Since no merge happened I changed the header. At the end it is best to be as general as possible.

bast:
I have only one question: it seems two new files were introduced - why was this done? (there may be a very good reason, just trying to understand why existing files were not changed instead)

Apart from that looks good.

heikef:
The new files are my own scripts. I shared a while ago very similar ones but they only work for python 2. Now I added also some for python 3. The "old" ones should of course be deleted at one point but at the moment most users use python 2 for gimic. I can of course rename / delete them but then we must agree that gimic shifts completely to python 3. This is something the developer team or at least two of us must agree on together. What do you think?

bast:
Well, Python 2 has 8 months left of life: https://pythonclock.org.
So IMO we want to support Python 3 and drop Python 2 ASAP.

I am OK if the above is merged. I would have perhaps adapted existing files instead of adding new ones but no problem for me to carry the old files for a while

heikef:
That is all I need to hear. I will remove the old files and only keep the new ones. By now we decide to support python 3. Give  me 10 min for the next pull request, please.

bast (approved):
There are some debug prints left outcommented. But I don't mind if you don't mind :-)

heikef:
Removed debug output and found some print statements that needed to be adapted. If you ask me now if I want to squash the commit messages the answer is no. :)

heikef:
By the way I know the scripts have some flaws, I think @lnw will spot them quickly and suggest a good solution during a coffee break.

bast:
My understanding is that the PR is ready to go - merging.

heikef:
Thank you! Finally we managed. ;)

PR #234, opened 2019-04-14
shared conversion script for cml file courtesy of M. Dimitrova

heikef:
Script converts a file mol.cml (Angstrøm) to mol-bohr.cml (Bohr). After reading an online issue discussion with a new user I realized this script might be useful for some user. Alternatively open babel can be uses as described in the documentation.


bast:
Nice script!

PR #237, opened 2019-04-14
added paraview state file for vector plot

heikef:
There was some confusion online about paraview vector plots so I included a vector plot with a slice into our 3D example. The reason for this is that the user has here the option to move the plane / slice and learn more about the currents in the molecule.


heikef:
I do not understand why the change in requirements.txt appeared here. It was done on an extra branch.

heikef:
squashed the commit messages

bast:
I am a bit worried that we are adding a number of pretty large files into the repo. The examples directory contains 14 MB. Is it possible to make the examples smaller but still meaningful?

heikef:
The mol-bohr.cml file is needed and also the state file. Strange I have only 9.8 MB in total for the whole 3D directory.

heikef:
We could remove the 2D directory since we do not support the *.txt files any more. It has only historic relevance. What do you think?

bast:
The 3D one is the big one. Is there a way to reduce those? Smaller grid, etc.?

heikef:
For demonstrating plots reducing the grid is not a good idea. The 2D directory is not needed in my opinion.

bast:
If we need huge grids, perhaps we can put them into a separate repo under https://github.com/qmcurrents/? Whatever we add into this repo here, we can never again remove unless we rewrite the history.

bast:
It is OK for me to remove 2D but this will not solve this problem.

heikef:
We will not extend this example. I wanted to share the paraview.pvsm file with a new user who is struggling a lot.

bast:
> We will not extend this example. 

Famous last words :-)

I will merge but it is super important that we watch the size of files.

heikef:
I understand and will keep this in mind. We have to write a tutorial to help new users, but this will be part of the documentation.

bast:
I agree about the tutorial. But this does not necessitate having huge files in this repo. Anyway, merging.

heikef:
Ideally we will delete this folder at one point. Thank you for merging.

bast:
Yes, but we will not get rid of its size. Everything that we merge we can in principle never remove.

heikef:
Thanks for raising awareness.

PR #240, opened 2019-04-21
Feature numgrid

### LINKED ISSUE #33: [Feature] link numgrid library to gimic
Therefore we need to modify the cmake stuff. Numgrid is an external library generating DFT grid files. https://github.com/dftlibs/numgrid

heikef:
* wrote a python script that reads in MOL and generates the needed input data for the numgrid library
* the generated grid is written out as well as information about the weigths and coordinates in bohr
* then gimic is run using the cdens module and the feature of reading in an external grid
* a new keyword called prop is defined in section essential - it can be switchen on and off
* prop=on generates for each atom the isotropic shielding constant and the magnetic suceptibility / magnetizability for the whole molecule 
* I encountered the well known sign problem that is hidden somewhere in gimic because only with negative fields I got positive shieldings and negative magnetizability - benzene from the examples directory was my test example 
* no documentation is provided yet since this is a first version of a new feature and thus not recommended for common users 
* closes #33 
* I used this type of input file: 

# NEW INPUT

calc=cdens          # cdens, integral, edens, divj
title=""
basis="MOL"
xdens="XDENS"
debug=1           # debug print level
openshell=false
magnet=[0.0, 0.0, 1.0]

# Read in grid from external file. 
Grid(file) {   
     file=gridfile.grd
}

Advanced {
    lip_order=5     # order of Lagrange Interpolating Polynominals to use
    spherical=off   # don't touch, unless you REALLY know what you are doing
    diamag=on       # turn on/off diamagnetic contributions
    paramag=on      # turn on/off paramagnetic contributions
    GIAO=on         # turn on/off GIAOs. Don't change unless you know why.
    screening=on    # use screening to speed up
    screening_thrs=1.d-8   # Screening threshold
}

Essential {
    prop=on  # makes only sense when numgrid is integrated
    acid=off
    jmod=off
}


heikef:
@lnw you are of course also welcome to comment

heikef:
* on the downside: with this new grid no ParaView plotting is feasible because we need an even grid for this 
* on the other hand when we are only interested in quick 3D property integrations it does not matter
* I switched the *.vti print function off when the keyword "prop" is set

heikef:
On my computer all tests passed. I think this must be related to python 2 and python 3 on TravisCI. At the moment I do not see what I might have overseen.... 
It is a numgrid installation bug. 

I could remove numgrid from requirements.txt then the test will pass....

bast:
Travis complains about the same problem which you have seen locally and which I have fixed on a branch of Numgrid but I have not yet merged this and not yet updated the package on PyPI.

heikef:
Yes, therefore I removed it from requirements.txt and now the tests run through.

bast:
Before we merge this I will need to fix the numgrid build problems. I am working on it. Also it would be good to have a test case for this new functionality. Later we will want to move the grid generation deeper into the code and not run it as an extra step and then the test case will really help.

heikef:
I am not sure on how to set up a test case for this. At the moment we run a script first and then gimic. I would like to discuss and optimize the code first and later on add a test example. As it is the code is not ready for common users. When we merge it will not cause harm.

bast:
But it will be difficult to optimize (without breaking it) if there is no test case. The minimum is a readme file which describes what to do and contains numbers that should come out. Later we can convert it to a script. If you can express what you do in a script, it is easy to add it to the test case.

src/fgimic/jfield.f90
```diff
@@ -522,6 +527,101 @@ subroutine acid_vtkplot(this)
...
+        natoms = getnlines(15)
+        allocate(coord(natoms,3))
+        do i=1, natoms
+          read(15,*) coord(i,1:3)
+        end do
+        close(15)
+
+        ! get weights from numgrid files (external)
+        open(unit=15, file="grid_w.grd")
+        open(unit=16, file="gridfile.grd")
```
bast, line 554: should we not put both into the same file instead of two files?

tools/NumGrid/mk_grid.py
```diff
@@ -0,0 +1,160 @@
...
+# print("alpa_max", alpha_max)
+# print("alpa_min", alpha_min)
+# print("")
+# print("alpha_min for center 1", alpha_min[0])
+# print("")
+# print("alpha_min for center 7", alpha_min[6])
+# print(x_coordinates_bohr, y_coordinates_bohr, z_coordinates_bohr)
+# print("max L", max_l_quantum_numbers) 
+fout = "coord.au"
+f1 = open(fout,"w")
```
bast: instead of open/close it is probably nicer to use `with open('coord.au', 'r') as f1`

tools/NumGrid/mk_grid.py
```diff
@@ -0,0 +1,160 @@
...
+
+    for k in range(num_points):
+        f1.write(str(x[k]) + " " + str(y[k]) + " " + str(z[k]) + "\n")
+        f2.write(str(w[k]) + "\n")
+
+    # num_radial_points = numgrid.get_num_radial_grid_points(context)
+    # print("number of radial points", num_radial_points)
+
+    # generate an isolated radial grid
+    # r, w = numgrid.get_radial_grid(context)
```
bast: we should also remove all the out-commented unused code

src/fgimic/jfield.f90
heikef: No. At the moment gimic reads in the grid file as xyz information. We can do this later on. The ordering is important to make sure that the J tensor is assigned correctly.

tools/NumGrid/mk_grid.py
heikef: Yes, I know. I tested if it works. If we keep it it would be easy to adapt the script, but we can also remove it.

tools/NumGrid/mk_grid.py
heikef: we want to write something should it not be "w" instead of "r"?

heikef: otherwise thanks for the hint, it saves writing a couple of lines

heikef:
I added the gimic.inp file as well as a README file containing reference values.

tools/NumGrid/mk_grid.py
heikef: Ok, cleaned up.

heikef:
I know the solution for numgrid as we have it here is not optimal but it is a start. Would it be acceptable to merge?

tools/NumGrid/mk_grid.py
bast: Yes. "w" for writing.

src/fgimic/jfield.f90
bast: I think one file is really better than two. But I understand if you don't want to change it, then I can change it later.

tools/NumGrid/mk_grid.py
bast: It is also "safer" since the file gets closed even if the code crashes half way so it's not just to save the extra line.

bast:
I plan to fix numgrid within the following days. It is OK for me to merge now and fix things later.

heikef:
All right. Then we will merge and after you have fixed numgrid we will add numgrid into the requirements.txt - I think this is an acceptable solution. 
It would be good to have the option of 3D integration at least rudimentary in the new release. Then we are also citable.

tools/NumGrid/mk_grid.py
heikef: ok.

src/fgimic/jfield.f90
heikef: I completely agree but we can adapt this later on. At the moment it would only complicate things. I tried to touch as little as possible inside gimic since we are after a completely different python oriented solution.

bast: OK - convinced :-)

bast:
You can also merge and I edit the `requirements.txt` and do small changes to your code in a follow-up PR.

heikef:
Ok, then I will merge my own pull request.

PR #244, opened 2019-05-02
added citation information into README.md closes #235

### LINKED ISSUE #235: add a citation section into README
I think adding the standard citations for using gimic would be helpful for new users. This should be done on the main / front page of our project.

heikef:
Added some relevant citations regarding the usage of GIMIC. Included also the G09 interface by Vincent because a lot of applications would not have been possible without it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,4 +42,7 @@ ACID:
 
 * H. Fliegl, J. Jusélius, D. Sundholm, Gauge-Origin Independent Calculations of the Anisotropy of the Magnetically Induced Current Densities, J. Phys. Chem. A, 120, 2016, 5658-5664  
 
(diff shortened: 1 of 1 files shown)


heikef:
For some reason the Valgrind test failed. I touched only the README.md file. It is unclear to me why we run into this problem.

heikef:
Would it be ok to merge anyway? The problem is for sure not the README.md file

bast:
The test is brittle. We need to fix the Valgrind test (#245). It is OK for me if we fix this later.

bast (approved):

heikef:
Ok, thank you. Then I will merge.

PR #248, opened 2019-05-22
Bugfixes and improvements to the bash scripts

mariavd:
I have made added some fixes and made some changes to the scripts. Little changes from the user's point of view, however there were some issues, so I think it is reasonable to merge the updates with the upstream.

Current profile re-plotted when crit_pts is called

Fixed: the y axis on the plot is divided by slice width

The y axis, not the x axis should be scaled by delta

typo in the plotting command

Renamed the parameter MF to B in the scripts; another way of calculating the guess for the fixed point


mariavd:
Actually, I just made an extra change. The labels on the current profile are now "Positive" and "Negative". We know that integrating across a molecular ring the paratropic current appears first in the negative and then in the positive contribution.

heikef:
Thank you!

PR #249, opened 2019-05-26
fix vti file export

lnw:
a cell is a volume and not a point.


heikef:
Thanks for fixing a bug! You changed several things. Could you please comment. Did you test? Do all visualization feature remain the same?

heikef:
The checks did not pass all. Perhaps you need to adapt the test suite?

lnw:
The bug was a misinterpretation of 'cells' in the vti file format.  A cell is a volume, defined by a set of vertices.  So we should be writing down a (scalar) value that corresponds to the volume.  Until now, there was just one value per vertex. So, the values were a) shifted by h/2 and b) there were too many (l*m*n instead of (l-1)*(n-1)*(m-1)).
 
The fix is to assign each cell the average of the eight corners.  That's not perfect, some higher order interpolation scheme would be better, but at least it's not silly-wrong.

lnw:
The remaining job fails.  I cannot fix that because I have no idea what it is supposed to do.  According to to jobdescription (read file), one would expect an arbitrary grid. gridfile.grd has rather randomly distributed points in a plane, that fits.  We never had a routine for writing vti files for arbitrary grids, and indeed the reference vti contains 'ImageData', ie a regular grid.  Furthermore the definition of 'cells' is not straightforward or even implicit for arbitrary grids.

a)  What was this supposed to be?
b) can we just delete+forget the vti file?

heikef:
This seems to affect only the norm. Maybe we should check if ParaView calculates automatically the norm of the vectors. Then we could delete this part and there would be no need to think about the averages. The definition is cell which is perhaps not correct but what we visualize is the vector at a certain point.

lnw:
Is that a reply to the whole patch? --> at least most visualisations don't use the cell-field, but i don't know if any use it.  Leaving it empty doesn't work.  Setting all to zero is possible, but then one might just as well write something useful.

Or does your comment refer to C4H4? --> The grid of that job doesn't make sense either way.

bast:
Looking at the history it seems the test was added to solve #97 by Heike. I know too little about VTI functionality to judge whether we should keep or delete it. Heike, what is your verdict?

heikef:
I will have a closer look soon.

heikef:
This test is only to read in an external grid file. An vti file is created which serves as reference. When changing the vti output one needs to adapt the respective reference file.

lnw:
No.  This test should not write a vti file (and I have changed the code such that it doesn't).  The explanation is, that vti is a file format for regular grids (aka ImageData).  When we read a gridfile we must expect any kind of points.  Those cannot be written using the vti file format.  They could be written as vtu (for example) but we don't support that at the moment.  So, at the moment, the only sane solution is to not print a grid.

heikef:
Yes, when we read in an unstructured grid the vti file does not make sense, because we can not visualize the data, but we can compare the calculated results for J, which is part of the vti file. What would be a good reference otherwise? We use the cdens module. I think the integral module does not work with an unstructured grid at the moment. 
As it is your modified output file does not make sense as a reference either.

src/fgimic/jfield.f90
```diff
@@ -376,8 +376,8 @@ subroutine jvector_plots(this,tag)
         end if
 
         ! put jvec information on vti file
-        if (this%grid%gauss.or.settings%prop) then
```
heikef, line 379: Why did you change this line?

lnw: Because what keeps us from printing a vti is not the properties but the externally created grid.  If used correctly, the externally generated grid comes together with the request for properties, but at the moment it's possible to request properties with a normal grid, and receive a bunch of NaNs.

lnw:
No, the problem is not that we cannot visualise the vti file. The problem is that the vti file would contain data that does not represent our system, not even close.  You're basically suggesting to use the jvec.vti as a 'hash': no intrinsic meaning but same input leads to same output.  I'm very much against that.

src/fgimic/jfield.f90
heikef: Why did you not report this when filing the pull request? The feature "properties" has very recently been added, there is no test available yet.

heikef:
I agree the solution is not optimal but at least we would test if the values for the grid points were calculated correctly.

heikef:
A better solution would be to figure out on how the visualization for unstructured grid works. You mentioned the vtu file format and adapt the program.

lnw:
Setting longterm wishes aside, the question is, whether there should be some nonsense code which allows for a real feature to be tested.  For me the answer is 'no' (and I'm aware that the correctness of the read grid is untested, but we do test that there are no errors).

src/fgimic/jfield.f90
lnw: It's just too much work to explain every line of code.

heikef: It was one line of code.

heikef:
1) The code setting up the vtk visualization for even grids is not a nonsense code. 

2) The situation for this specific test case is that the read grid option is not properly tested. I think we both agree on this. Do you have a suggestion on how to test it properly? As I understood - no. You pointed out that it is not a good idea to do this by checkting the created vti file, which is fine. Instead you suggest to test only that the program does not crash. Now we have to decide if this is sufficient or if there is a way to improve and test properly. 

3) Long term wishes arising from this discussion will lead to the creation of a new issue and improve the code in the long run. We should figure out on how to improve the read grid test. 

4) Did you check if the calculation of the norm in the vti file is needed of if it can be skipped? If not and if you do not wish to do this (which is perfectly fine with me) please tell. Then we should create a new issue to make sure that we keep this on the radar. 

5) I think we do not need 6 commit messages. Would it be ok for you that I squash them or do you prefer to do this yourself? A question as commit message is not meaningful in my view.

bast:
As to the commit message granularity: I think this is fine except perhaps the merge commit.

bast:
I gathered from this back and forth discussion that we wish to test the reading of a grid and that we have done that by comparing an output file? Should we not rather test a computed result for a reference grid or a series of reference grids to test the grid reading?

heikef:
The computed result was part of the vti file, which should not be generated for an unstructured grid. Lukas is right in this point. A possible compromise would be to do the merge but to create an issue that we remember to improve the test and check if the norm calculation is really needed in the vti file.

bast:
Yes, I also upvote merge and follow-up issue.

lnw:
As for 1) I was calling the usage of the regular grid export for non-regular grids non-sense; certainly not the regular grid export per se. 2) I think just reading a grid is a very simple operation which needs not be tested at all cost.  4) As I was writing in my fourth post in this thread: Most paraview visualisations we use don't read the cell data but there has to be something. So I would write norms rather than filling it with zeros. 5) sure, squash everything.

heikef:
Ok, I think we converged to a solution. We merge and fill a new follow up issue. Thanks for your  contribution Lukas! @lnw

Issue #250, opened 2019-05-26
Adapt vtk output to unstructured grid


heikef:
As it is we assume an even grid for the vtk visualization in ParaView. It would be good to include also unstructured grids. Any idea on how to achive this? @lnw had some ideas when I remember correctly

heikef:
vtu file format might be a solution for this

PR #254, opened 2019-06-04
enable unstructured grid printing, test

lnw:
feel free to squash


lnw:
as for the failing test ... what is the point of choosing cdens + bond?

heikef:
A bit more explanation about what you do and why would be nice.

heikef:
Why is the bash script under tools added?

heikef:
The option to use cdens+bond is important. A jmod.txt file should be created then, which can be visualized. This is very useful when doing a quick test for setting up the dimensions of the integration plane. I would like to keep it and removing it is not an option for me.

heikef:
A jpg file with a screenshot of the vtu visualization would also be useful to have.

heikef:
Otherwise it is great that you took initiative to improve the test and the visualization tools.

heikef:
This would close issue #250

lnw:
@heikef  There will be another commit some time with an explanation of how to actually do this.  The extremely short bash script will become obvious then, it's for input preparation.

What do you want to do with a jpg?  It just looks like a normal paraview visualisation.

I'll see how I can restore the cdens+bond.

heikef:
A jpg file would document how the output is expected to look like. If somebody never used ParaView it will not be obvious for him/her.

heikef:
Always imagine the worst possible user. For this guy the bash script might not be obvious. A short comment for documentation will be helpful in case we need to trace things back.

lnw:
Hah!  How do you know the worst possible user is a guy?

lnw:
@heikef @mariavd I have written some instruction on how to do the unstructured grid stuff.  Have a look, it should be fully explanatory.

heikef:
@lnw - guy was meant gender neutral - next time I will use the term person to be more clear

heikef:
@lnw Thank you for your efforts! From my side the merge can be done and as I understood it is ok with you that I squash your commits. So I will wait now for @mariavd to share her opinion.

mariavd:
I promise to test the functionality today after the CSC course is over. I am sorry for being so slow as usual.

mariavd:
Sorry, I am feeling stupid. I saw the small bash script to generate the grid but what do I actually give as an argument to it?

bast:
> I promise to test the functionality today after the CSC course is over. I am sorry for being so slow as usual.

No you are not slow. You give a lot to this project for free.

doc/visualization.rst
```diff
@@ -2,11 +2,42 @@
...
+interpolates the values given at grid points within each cell.
+
+In case of vti files, all required information is given by the grid and no
+further input is required.
+
+In case of vtu files we require an external program to create a tetrahedral
+mesh.  One choice for such a program is TetGen
+(http://wias-berlin.de/software/index.jsp?id=TetGen&lang=1).  The workflow is:
+
+1. obtain a grid of your choice.  The file format should be: no header entries,
```
mariavd, line 29: @lnw How do I choose the grid of my choice?

mariavd:
And if you think it might be useful to have an alternative 3D-run.sh or maybe include the choice of grid as a question to the user, it will take little effort to prepare it. 

@bast Thank you. It is part of my PhD, so I feel obliged to contribute. I have various ideas for making gimic more user-friendly but many might not get done. At least more documentation notes will be on the way.

heikef:
@mariavd Thanks for testing!

doc/visualization.rst
lnw: I hate to repeat myself, but that's your choice :)

You could use numgrid to obtain something spherical around each atom (but that's a poor/expensive choice for visualisation).  I think the typical usecase is to investigate an oddly shaped molecule and write a custom grid for that.  We could for example have done that for the tori where just the surface +/- a bit is required.

In general, any set of points will do, You could even randomly sample points in whatever volume.

heikef: It just came to my mind if we have the surface around a molecule we could also do a similar visualization like the ACID community does, namely plotting the current density vector on top of the ACID surface chosen at a specific isovalue.

heikef:
I will squash and merge now. Thanks to all of you for your contributions!

doc/visualization.rst
lnw: That doesn't sound very thaight forward ... we don't have an ACID isofurface, right?  Can paraview extract isosurfaces from whatever grid (seems plausible)?  If so, one could assign ACID to cells and the rest is probably easy.

lnw:
Before I add more semi-related thing to this pull request ... :)

doc/visualization.rst
heikef: For an even grid we have the ACID values. A file called acid.vti is created and with Paraview one
can visualize the isosurface. If we have a grid only covering the surface of a molecule it should be possible to visualize both the acid surface and the current density vectors. Otherwise - yes, it is a side discussion and not further relevant for this pull request. We can decide if we want to fill an issue.

PR #265, opened 2019-07-25
fixed test issue closes #261

### LINKED ISSUE #261: benzene/3d and benzene/3d-keyword-magnet tests failed
For some reason these tests fail after updating my gimic version. I got suspicious and did a complete new clone and the problem remains. Could it be that we need to set a tolerance
delta x for checking because depending on the machines and operating systems the numerics might not be identical? 
On another machine I had the same problem. Therefore I think it is something more serious 
and not related to my local setup. 
Could someone please test and tell if he/she has the same problems? Thank you! 
It is strange that all tests passed on the TravisCI though....

heikef:
Now all tests pass in a reasonable amount of time.

### DIFF
--- a/test/benzene/2d-keyword-magnet/test
+++ b/test/benzene/2d-keyword-magnet/test
@@ -12,7 +12,7 @@ assert version_info.major == 2
 options = cli()
 
 # we check entire files
-f = [get_filter(rel_tolerance=1.0e-8)]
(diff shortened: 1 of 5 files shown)


bast (approved):
Yay! Thanks.

heikef:
I have to thank you, because you fixed the issue in fact.

PR #266, opened 2019-07-30
check if some files even exist, before reading.  …

lnw:
Reduces the number of confused users.

### DIFF
--- a/src/fgimic/jfield.f90
+++ b/src/fgimic/jfield.f90
@@ -272,6 +272,7 @@ contains
         real(DP) :: bound, r
         integer(int32) :: dummy, ncells
         logical                            :: fd2_isopen = .false.
+        logical                            :: elements_exists
 
         if (mpi_rank > 0) return
         ! get rid of print out of txt files but keep them for debugging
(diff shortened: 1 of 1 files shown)


bast:
Change looks good. Just to confirm: in this case it makes sense to not stop the calculation if these files do not exist?

lnw:
For a missing ele it definitely makes sense to not abort.  In case of the other three, it would as far as I can see be adequate to abort.  Runtime wise it doesn't matter, because not much else happens.  If we ever get the option to do several tasks in a single job file, it would make sense to bail out selectively.
So, I slightly tend to leave everything as it is, but I can also abort.

bast:
These are generated files? If yes, how can they not exist?

I wonder whether according to the principle of minimum surprise it would not be more careful to stop with a nice error message. I am always worried about producing strange numbers. Crash is better than wrong numbers IMO.

lnw:
The files are generated with a) a script in tools and b) a third party prog.  It is the user's responsibility to create them and a user could fail to do so.

This is exactly why I'm printing a message saying that files are missing and therefore nothing will be done. Previously there were wrong (=0) results.

bast:
OK - thanks for explanation!

lnw:
Tak skal du ha!

PR #269, opened 2020-04-15
added SI unit conversion to magnetizability calculation

heikef:
added a unit conversion to the magnetizability section to SI units, this will make a comparison to other work easier


bast (approved):
Thank you!

PR #270, opened 2020-10-10
worked on integrand visualization

heikef:
Worked on integrating visualizations for unstructured grids taking the work of @lnw as example. However, the surface pictures look bad and I do not understand why. Now it is time to share and ask for comments.

### DIFF
--- a/src/fgimic/jfield.f90
+++ b/src/fgimic/jfield.f90
@@ -582,16 +582,24 @@ contains
     end subroutine
 
     subroutine get_property(this)
+      use vtkplot_module
         ! assume external grid is read in correctly via read grid
         ! assume grid_w.grd, coord.au, gridfile.grd are in the same directory
+        ! for visualization of unstructured grid file grid.1.ele is needed
+        ! assume this has been generated with tetgen and is in the same directory
         type(jfield_t) :: this
         real(DP), dimension(:,:), pointer :: jtens
         real(DP), allocatable :: wg(:), coord(:,:), grd(:,:)
+        real(DP), allocatable :: intchi(:,:), pdata(:), intsigma(:,:)  
         real(DP), dimension(3) :: d, bb, jvec, sigma, chi
         real(DP) :: f, tmp, si_tmp
-        integer :: i, j, k, npts, natoms
+        integer :: i, j, k, npts, natoms, dummy, ncells
+        integer(int32), allocatable :: cells(:,:) ! for vtu file
         logical :: coords_exists, points_exists, weights_exists
+        logical :: elements_exists
         real(DP), parameter :: fac_au2simag = 7.89104d-29
+        character(len=70) :: filename
+        character(len=70) :: formatstring
 
         inquire(FILE='coord.au',     EXIST=coords_exists)
         inquire(FILE='gridfile.grd', EXIST=points_exists)
@@ -627,12 +635,25 @@ contains
         close(15)
         close(16)
 
+        ! get cell information for printing vtu file
+        inquire(FILE='grid.1.ele', EXIST=elements_exists)
+        if(elements_exists) then
+          open(GRIDELE, file='grid.1.ele')
+          read(GRIDELE, *) ncells, dummy, dummy
+          allocate(cells(4,ncells))
+          do i=1,ncells
+            read(GRIDELE, *) dummy, cells(1:4, i)
+          end do
+          close(GRIDELE)
+        end if 
+
         ! assume jtens has been calculated in the same order
-        ! loope atoms
+        ! calculate shielding tensor diagonal elements
+        allocate(intsigma(npts,3),pdata(npts))
+        ! loop  atoms
         do k=1, natoms
           ! loop grid points
           sigma = 0.0d0
-          chi = 0.0d0
           do i=1, npts
             ! loop xyz
             do j = 1, 3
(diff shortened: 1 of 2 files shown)


lnw:
do you have an example for what goes wrong?

src/fgimic/vtkplot.f90
```diff
@@ -311,5 +311,83 @@ subroutine write_vtk_vector_unstructuredgrid(fname, grid, pdata, cells)
         call closefd(fd)
     end subroutine
 
+    ! fname: output filename
+    ! grid: grid point coordinates
+    ! pdata: vectors at grid points
```
lnw, line 316: those are now scalars at grid points

src/fgimic/vtkplot.f90
```diff
@@ -311,5 +311,83 @@ subroutine write_vtk_vector_unstructuredgrid(fname, grid, pdata, cells)
...
+    ! cell definitions
+    subroutine write_vtk_imagedata_unstructuredgrid(fname, grid, pdata, cells)
```
lnw: I think I'd call this 'write_vtk_scalar_unstructuredgrid'.

src/fgimic/jfield.f90
```diff
@@ -674,6 +701,66 @@ subroutine get_property(this)
           tmp = (sigma(1) + sigma(2) + sigma(3))/3.0d0
           write(*,*) "shielding constant sigma in ppm= ", tmp
 
+          if(elements_exists) then
+              if (k < 10) then
```
lnw: that's one level of indentation too much

src/fgimic/jfield.f90
```diff
@@ -627,12 +635,29 @@ subroutine get_property(this)
...
+        ! read element file: first number in first line is number of lines
+        inquire(FILE='grid.1.ele', EXIST=elements_exists)
+        if(elements_exists) then
+          open(GRIDELE, file='grid.1.ele')
+          ! read(GRIDELE, '(3i4)') ncells, dummy, dummy
+          read(GRIDELE, *) ncells, dummy, dummy
+          ! write(*,*) "ncells ", ncells
+          allocate(cells(4,ncells))
+          do i=1,ncells
+            ! read(GRIDELE,'(5i7)') dummy, cells(1:4, i)
```
lnw: i'd probably delete a few outcommented lines here and further down

src/fgimic/jfield.f90
```diff
@@ -674,6 +701,66 @@ subroutine get_property(this)
...
+              call write_vtk_imagedata_unstructuredgrid(trim(filename), this%grid%xdata, intsigma(:,2), cells)
+              write (filename,formatstring) "sigma_zz", k, ".vtu"
+              print *, filename 
+              call write_vtk_imagedata_unstructuredgrid(trim(filename), this%grid%xdata, intsigma(:,3), cells)
+          end if
+        end do !loop atoms
+        write(*,*) ""
+        
+        ! calculate diagonal elements of magnetizabilitz tensor
+        ! for magnetizabilitz no loop over coords is needed
```
lnw: someone has a German keyboard layout :)

src/fgimic/vtkplot.f90
heikef: but you are in the vector file

src/fgimic/vtkplot.f90
heikef: hm, this is a matter of taste. For the structured grid we called it also "image data".

src/fgimic/jfield.f90
heikef: ok

src/fgimic/jfield.f90
heikef: Ok. When everything works we can do a clean up.

src/fgimic/jfield.f90
heikef: Nope, a Norwegian one.

heikef:
@lnw Thanks for commenting and reviewing! Despite of the name of the vtu scalar print file you did not comment further. I assume the modification was correct then. When removing the integral weights the visuals looked also much better.

heikef:
@lnw and @bast when you are ok now with the changes I would suggest to merge and then the decomposition into positive and negative contributions from Rahul @rahulj98 could be added

src/fgimic/vtkplot.f90
lnw: to me, we are printing a scalar field (which also applies to the comment one up).  This is in analogy to the vector field that is printed in the other function that is called 'write_vtk_vector_unstructuredgrid'.

lnw:
I can't comment on the correctness of the equations, such as the factor of 1/3, but I did verify that the printed scalar fields are now smooth.  Apart from that it looks very useful.

src/fgimic/vtkplot.f90
heikef: I changed the name in my last commit.

heikef:
@lnw thank you for the verification. I removed the factor of 1/3 because then one can use larger isovalues for visualization and the resulting plots remain the same. Thus, the factor does not change the visual information (at least for c4h4 and h2o).

heikef:
Cleaned up comments and squashed commits.

heikef:
@bast I would merge now. Then Rauhl could contintue and file a new pull request.

PR #271, opened 2020-10-12
positive and negative contribution of the shielding constant

rahulj98:
added a way to calculate the positive and negative contributions to the shielding constant by calculating the shielding contribution for each point.
also, we can now create the jvec.vtu files for each atom in a molecule.

### DIFF
--- a/src/fgimic/jfield.f90
+++ b/src/fgimic/jfield.f90
@@ -24,6 +24,7 @@ module jfield_class
         real(DP), dimension(3) :: b
         real(DP), dimension(:,:), pointer :: tens
         real(DP), dimension(:,:), pointer :: vec
+        real(DP), dimension(:,:), allocatable  :: sigmaval
         type(grid_t), pointer :: grid
     end type
 
@@ -367,6 +368,7 @@ contains
         else if ( ((trim(this%grid%mode)=='base' .or. trim(this%grid%mode)=='std') .and. this%grid%gauss) &
               .or. trim(this%grid%mode)=='file' ) then ! (base && gauss) || file
           allocate(jval_unstructured(this%grid%npts(1), 3))
+          allocate(this%sigmaval(this%grid%npts(1), 3))
           dummy = 0
           do i=1, p1 ! which is the total number of points
             v = jv(:, i)
@@ -374,6 +376,7 @@ contains
               coord=gridpoint(this%grid, i, dummy, dummy)*AU2A
               call write_jmod(coord,v,fd2)
             end if
+            this%sigmaval(i , 1:3) = v
             jval_unstructured(i, 1:3) = v
           end do
         end if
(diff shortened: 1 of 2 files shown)


heikef:
I think we should write out the shielding density as a density and not as a vector file. Writing a vtu file for each atom is the same thing as done in my pull request. 
We should keep the separation in positive and negative contributions. 
There are too many commits. Please squash.

src/fgimic/jfield.f90
```diff
@@ -367,6 +367,7 @@ subroutine jvector_plots(this,tag)
         else if ( ((trim(this%grid%mode)=='base' .or. trim(this%grid%mode)=='std') .and. this%grid%gauss) &
               .or. trim(this%grid%mode)=='file' ) then ! (base && gauss) || file
           allocate(jval_unstructured(this%grid%npts(1), 3))
+          allocate(this%sigmaval(this%grid%npts(1), 3))
```
heikef, line 370: Why did you add this to the type structure for the J tensor? Outside the property routine this entry will not be needed.

src/fgimic/jfield.f90
```diff
@@ -425,19 +421,29 @@ subroutine jvector_plots(this,tag)
             ! read(GRIDELE, '(3i4)') ncells, dummy, dummy
             read(GRIDELE, *) ncells, dummy, dummy
             ! write(*,*) "ncells ", ncells
-            allocate(cells(4,ncells))
+            allocate( this%grid%cells(4,ncells))
```
heikef, line 424: Here the inclusion of the cells makes sense. On the other hand we only need this when reading in grid.1.ele.....

src/fgimic/jfield.f90
```diff
@@ -425,19 +421,29 @@ subroutine jvector_plots(this,tag)
...
+            deallocate(jval_unstructured)
+            deallocate(this%sigmaval)
+            deallocate( this%grid%cells)
+
           else
             write(*,*) 'not writing a vtu file, because the file grid.1.ele was not found.'
           endif
-          deallocate(jval_unstructured)
+
+          ! Maria: moved this call here since the cells are needed when printing
```
heikef, line 444: This comments are not needed later on.

src/fgimic/jfield.f90
rahulj98: I am sorry if this is wrong I have limited knowledge of Fortran. 
The variable sigmaval was defined at line 27 in the data type jfield_t, which was called in the subroutine jvector_plots as 'this' at line 256

bast:
Thanks Rahul! Any way I can help to get this change in?

src/fgimic/jfield.f90
heikef: No need to be sorry. I suggest to avoid adding extra fiels to fix type structures. We need this only in the property routine.

heikef:
@bast - I filed a pull request first. When my pull request is merged Rauhl could put his stuff in. Instead of writing the tensor components as vector information on file we should use a new density type format ("scalar"). This has been done (see my pull request). After the merge Rauhul could add the collection of the positive and negative contributions to sigma.

PR #285, opened 2021-05-13
adapted script to new NumGrid API

heikef:
Adapted the grid generation script because the NumGrid API changed.


bast:
Thanks for the update!

PR #291, opened 2021-11-29
move testing from travis to github actions; closes #290

### LINKED ISSUE #290: Move testing from Travis to GitHub Actions
I have the impression that Travis does not run the tests anymore.

bast:


bast:
Merging. Without this we have no automated testing. With this we have it back so this is essential. Maybe I missed something but that can be ironed out in a follow-up PR.

PR #292, opened 2022-09-07
Apptainer container support

tjjarvinen:
This will add definition file to build [Apptainer](https://apptainer.org/) container, and a README.md file to give instructions on building and using the container.

The main point with this is to have easy access to Gimic. Basicly you only need to install Apptainer, after which you can build the container and have access to Gimic without the need to install any dependencies. Also, moving Gimic binary (container) to other computers becames trivial with the container framework.
 The choise of Apptainer (former Singularity) was taken because it is developed with HPC in mind and many supercomputers (like the ones in Finland) use it.

The problem with the container is that some of the scripts in `jobscripts` folder call external programs. These wont work with the container. Because by definition container does not depend on external programs. This mainly affects scripts with "cluster" in their name, as they call Slurm. The versions with "local" in their names should work.

For visualization scripts, I added gnuplot into the container. This does however increase the size of the container by an extra 100MB to about 370MB total.

@mariavd promised to edit the scripts, so that they work with the container.
 In my opinnion, the way to go, is to remove all Slurm commands from the scripts and have the scripts output Slurm batch file, which the user can submit outside the container. That being said there is a lot of room to improve the scripts in general...

But other than that, the Gimic container should be fully working.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,3 +14,5 @@ share/*
 cmake/bump-version
 venv/
 *.DS_Store
+
+*.sif
--- a/container/README.md
+++ b/container/README.md
@@ -1,0 +1,59 @@
+# Apptainer Container
+
+This is guide on [Apptainer](https://apptainer.org/) container use for Gimic.
+
+## Building Container
+
+Build container
+
+```bash
+cd  "gimic folder"
+apptainer build --fakeroot gimic.sif container/gimic-apptainer.def
+```
+
+**Note**, you need to be in gimic folder or the build will faill. Apptainer does not support relative paths to definition file, yet, so this is important!
+
+## Using Container
+
+Container file is an executable by on its own and
+is equivalent to `gimic` command.
+
+```bash
+gimic.sif gimic.inp > gimic.out
+```
+
+All commands can be called with `apptainer exec gimic.sif <command>` calls, e.g.
+
+```bash
+apptainer exec gimic.sif gimic gimic.inp > gimic.out
+apptainer exec gimic.sif 3D-run.sh
+```
+
+This includes all executables that are located in
+
+- `build/bin/`
+- `jobscripts/`
+- `tools/`
+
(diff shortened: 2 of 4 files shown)


heikef:
Dear Teemu @tjjarvinen, thank you for this initiative! I think the benefits are clear. May you please comment on why you did not go for a docker solution? I am asking this just out of curiosity. 
From my point of view it would be important that a) all test we have so far are running through and b) that this new approach is tested by several people. 
I am assigning @bast to review your changes since he knows best about HPC and containerization. Hope this is ok with you, Radovan @bast since it implies extra work.

heikef:
Nice! I see all tests passed through! No objections from my side to merge. Regarding the bash scripts we got feedback that users find them very helpful. Sooner or later we should get a python version of them. @mariavd  will for sure disapprove. ;) Regarding gnuplot, it is only helpful when you need checking the integrations planes by hand (eg. when neighboring currents are present) otherwise the integration scans are very reliable.

bast:
Thanks a lot! As to Apptainer vs Docker: Apptainer is more suitable for multi-user clusters (although one can convert Docker images) so I support that choice. I did not yet look at the changes but will do so ASAP.

bast (approved):
This looks great for me. Thanks so much for this contribution!

bast:
Oops sorry for a late question. Why is the density.txt removed as part of this PR?

tjjarvinen:
> Oops sorry for a late question. Why is the density.txt removed as part of this PR?

`desity.txt` is a symbolic link to a file that does not exist, which causes an error when Apptainer moves files to build the container.  So, to make container building work the broken symbolic link needed to be removed.

> May you please comment on why you did not go for a docker solution?

The main issue with Docker is that it runs containers in root space. Meaning that you need to have root access in order to run containers. Apptainer on the other hand runs containers in user space, which makes it much more usable in HPC environment.

bast:
Thank you for the explanation!

Issue #193, opened 2018-09-13
Even "drier" dry run


mariavd:
When one handles complicated molecules it is necessary to preview the integration plane, sometimes more than a couple times. This only requires the grid to be defined. In a huge molecule of some 800 atoms and more this means at least 15 minutes. Currently, gimic reads in both the basis and the density matrices and transforms the orbitals, and only afterwards allocates the grid. 

@lnw suggested a reasonable idea that an "extra dry" is made, where only the MOL data is read in and the density matrix data is not touched at all. If you approve of the idea, I can take care if it.

heikef:
Are information about the basis sets needed for the grid set up?

lnw:
The grid generation should require purely geometrical data only.
I was of course joking about the 'super-dry' run, maybe a useful argument name would be '--generate-grid'.  I would expect such a run to return in less than a second, anything else is not justified.

bast:
I agree, if grid generation takes more than 2 seconds, something is not right.

lnw:
So we agree, that a '--generate-grid' option which does nothing else would be helpful?

bast:
Yes I think this would be helpful.

heikef:
I think it would make more sense to replace the old "dryrun" option with this new option we are discussing. All a user needs when doing a dryrun is the placement of the integration plane, which
depends on the grid. Instead of introducing the "generate-grid" option we could keep the dryrun option and modify it.

PR #297, opened 2023-03-26
fixed cfour hash issue for two character elements

heikef:
- included a replacement of "#" to " #" in MOL file into mk_grid.py
- problem was that two character elements have not been recognized properly  like "NI#1" now "NI #193 
- note this holds only for MOL files generated by CFOUR

### DIFF
--- a/.github/workflows/test-debug.yml
+++ b/.github/workflows/test-debug.yml
@@ -11,7 +11,7 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        python-version: [3.8]
+        python-version: ["3.10"]
         os: [ubuntu-latest]
 
     steps:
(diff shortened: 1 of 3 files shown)


heikef:
@bast I changed only a script that does not influence the gimic build at all. Do you happen to have an idea why all tests are failing? 
- looks like a cmake problem: Could NOT find Python3 (missing: Python3_NumPy_INCLUDE_DIRS NumPy) (found version "3.10.6")

bast:
I am looking into it ...

bast:
That problem is now fixed. And another appeared. Something seems to leak memory. But this might now come from this PR.

heikef:
How can this be? I changed only a script used outside of gimic. Anyway, thank you for looking into this!

bast:
Sorry I meant to write that this probably does not come from this PR. I will merge. The memory leak problem is probably older and needs to be investigated separately.

